PR #32, opened 2018-07-26
fix: wrong boost libraries found at runtime; use static boost libraries

rffrancon:
I tested this on d80db6dd534ee4b48626e0b336a4eee9df1607ad which is a little while back. But I'm feeling confident that it will work with the latest dev branch commits.

When I tested it, all unit tests pass on Yoda passed. Seg fault during tests was fixed. The wrong boost shared libraries were used at runtime. I avoided that problem by linking against boost static libraries.

@mbhall88 please remove `/nfs/leia/research/iqbal/software/boost_1_62_0`. I think that source/build files should be here: `/nfs/leia/research/iqbal/software/tmp`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,8 @@ project(${PROJECT_NAME_STR} C CXX)
 
 find_package(Threads REQUIRED)
 find_package(ZLIB REQUIRED)
+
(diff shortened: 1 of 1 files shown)


mbhall88:
@rffrancon I get permission denied when trying to remove that directory for some reason.

PR #37, opened 2018-10-01
Index optimize

rffrancon:
Not ready for merging. Sharing current code state.

### DIFF
--- a/src/localPRG.cpp
+++ b/src/localPRG.cpp
@@ -14,8 +14,9 @@
 #define assert_msg(x) !(std::cerr << "Assertion failed: " << x << std::endl)
 
 
-LocalPRG::LocalPRG(uint32_t i, std::string n, std::string p) : next_id(0), buff(" "), next_site(5), id(i), name(n),
-                                                               seq(p), num_hits(2, 0) {
+LocalPRG::LocalPRG(uint32_t i, const std::string &n, const std::string &p) : next_id(0), buff(" "), next_site(5), id(i),
+                                                                             name(n),
+                                                                             seq(p), num_hits(2, 0) {
     vector<uint32_t> v;
     // avoid error if a prg contains only empty space as it's sequence
     if (seq.find_first_not_of("\t\n\v\f\r") != std::string::npos) {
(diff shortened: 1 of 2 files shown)


rmcolq:
I didn't know that using std:: was faster than importing std?! Looks good so far.

rffrancon:
To be clear, using `std::` shouldn't make the code any faster. It's basically linting.

src/localPRG.cpp
```diff
@@ -60,25 +60,37 @@ std::string LocalPRG::string_along_path(const std::vector<LocalNodePtr> &p) {
...
+            uint32_t node_end = node_ptr->pos.get_end();
+
+            bool is_within_node = interval_end > node_start and interval.start < node_end;
+            bool is_equal_node = interval.start == node_start and interval_end == node_end;
+            bool is_other = (
+                    interval.start == node_start
+                    and interval.length == 0
+                    and it == --(p.path.end())
+                    and n != prg.nodes.begin()
+            );
```
rffrancon: @rmcolq From what I understand, this code should be functionally identical.
What is `is_other`? What can I use as a better variable name?

rmcolq: `is_within_node` should be something like `is_overlapping_node`, and I guess `is_other` should be `is_null_node` because it is about allowing nodes of length 0, within the criteria.

rffrancon: Thanks.

rffrancon:
@rmcolq More optimizations are possible. Please review current changes. If all is well, I propose merging these changes before further work to this area of the code.

rmcolq:
I'd quite like to keep the comment 
```
// because the local nodes are labelled in order of occurance in the linear prg string, we don't need to search after this
```
on what was line 78, is now line 90. Every now and again we talk about generating PRGs a different way, or integrating Michael's stuff as a new site, and that comment will remind that there is code which makes the assumption that sites are numbered in order in the PRG. Otherwise, would be happy to merge.

rffrancon:
That comment was kept. It was re-worded and moved up a few lines. Would you like me to change the wording or its position?

rmcolq:
That's fine. Feel free to merge. I don't have laptop or internet for a few hours.

Rachel

PR #39, opened 2018-10-06
Bring dev branches up to date across forks

mbhall88:
* Have added in the ability for local assembly generate multiple start and end kmers.  
* Added more logging
* Add some functions to calculate coverage on fastaq data structure instances


mbhall88:
All tests passing

Issue #68, opened 2018-10-18
Standardise the use of boost::filesystem for directories and filepaths


mbhall88:

mbhall88:
Given we want to standardise the use of boost filepaths, it maybe makes sense to standardise filestreams with boost also?  
There aren't necessarily any differences (boost just uses the STL fstreams), but boost's filestreams accept boost paths as arguments (at the moment you have to convert the boost path to a string in order to get fstreams to open it).  
https://www.boost.org/doc/libs/1_32_0/libs/filesystem/doc/index.htm

mbhall88:
I think the best way to "enforce" this throughout the codebase is if we parse the command line options that are paths as `fs::path` objects. It then becomes very apparent when you try to compile, which sections of the code need updating to boost filesystem.

leoisl:
Totally agree with this - we won't need to go scanning the codebase as to where we should update, the compiler will tell us! Nice idea!

PR #69, opened 2018-10-18
Standardise the use of boost::filesystem, add alternative covg to score function, and reimplement kmer coverage function

mbhall88:
#68


mbhall88:
This PR accidentally grew bigger than I had intended. It does four main things:
1. Adds some changes to move the project towards standardised use of `boost::filepath` for all paths - rather than using `std::string`. #68 
2. Adds an alternative coverage to score conversion function on the `fastaq` class. This new function maps a coverage (plus 33) to it's ASCII value. If the coverage is over the max value (93) then it just maps to the max ASCII value of `~`. This function is intended for getting coverage information on the consensus to assess coverage around the slices we cut out for local assembly.
3. Re-adds some function parameter defaults for local assembly. After we had refactored the code I forgot to re-add them.
4. Reimplement the expected kmer coverage function which was originally on the `fastaq` class. As we no longer use this class for capturing reads overlapping a slice for local assembly it has been rewritten as a standalone function.

PR #84, opened 2018-10-24
enhancement: disable logging when running unit tests

rffrancon:


rmcolq:
Fine with me if it's fine with @mbhall88 - that's all his logging work.

mbhall88:
Ah, I've been meaning to raise this but wasn't sure how to make it affect all tests. Good work!

mbhall88:
Only thing to keep in mind is that I dont think this will stop all the non-boost logging done in tests using just `std::cout`

Issue #23, opened 2018-06-11
Optimize index construction


rmcolq:

rmcolq:
Indexing the 23,000 genes takes 3 days 5 hours. Here are some excerpts from the log file for genes which took a long time
```
Wed Aug 29 15:09:20 2018 Sketch PRG GC00000205_10 which has 1000 nodes
Wed Aug 29 15:13:03 2018 Remove 'bad' edges from kmergraph

Wed Aug 29 15:51:42 2018 Sketch PRG GC00000027_6 which has 4088 nodes
Wed Aug 29 19:48:21 2018 Remove 'bad' edges from kmergraph

Wed Aug 29 22:07:03 2018 Sketch PRG GC00004221 which has 4572 nodes
Thu Aug 30 19:42:04 2018 Remove 'bad' edges from kmergraph

Thu Aug 30 19:43:45 2018 Sketch PRG GC00002479 which has 1287 nodes
Thu Aug 30 20:02:25 2018 Remove 'bad' edges from kmergraph

Thu Aug 30 20:54:11 2018 Sketch PRG GC00000222_4 which has 1537 nodes
Thu Aug 30 21:15:27 2018 Remove 'bad' edges from kmergraph

Thu Aug 30 21:15:42 2018 Sketch PRG GC00002914 which has 2470 nodes
Thu Aug 30 21:33:54 2018 Remove 'bad' edges from kmergraph

Thu Aug 30 22:24:38 2018 Sketch PRG GC00002737 which has 1978 nodes
Thu Aug 30 23:12:41 2018 Remove 'bad' edges from kmergraph

Fri Aug 31 23:56:52 2018 Sketch PRG GC00003373_r1_1 which has 4666 nodes
Sat Sep  1 01:22:40 2018 Remove 'bad' edges from kmergraph
```
Compared to fast times with medium numbers of nodes
```
Wed Aug 29 15:06:42 2018 Sketch PRG GC00007068_2 which has 221 nodes
Wed Aug 29 15:06:45 2018 Remove 'bad' edges from kmergraph

Wed Aug 29 15:06:46 2018 Sketch PRG GC00000085_5 which has 288 nodes
Wed Aug 29 15:06:51 2018 Remove 'bad' edges from kmergraph

Wed Aug 29 15:06:59 2018 Sketch PRG GC00001278 which has 656 nodes
Wed Aug 29 15:08:26 2018 Remove 'bad' edges from kmergraph

[...]

PR #88, opened 2018-10-29
WIP: enhancement #23: interval end memorized instead of recalculated

rffrancon:
Please don't merge. Looking for feedback.

Is there a reason why we cant just memorize the value of `get_end` for an interval? Is `Interval.start` ever changed after construction?

### DIFF
--- a/src/interval.cpp
+++ b/src/interval.cpp
@@ -6,15 +6,16 @@
 #define assert_msg(x) !(std::cerr << "Assertion failed: " << x << std::endl)
 
 
-Interval::Interval(uint32_t s, uint32_t e) : start(s) {
-    assert(e >= start); // not a real interval ;
-    assert (e - start < std::numeric_limits<uint32_t>::max() or
-            assert_msg("Interval [" << s << "," << e << ") was too long"));
-    length = e - start; // intervals need to be exclusive of end point so that empty strings can be represented
+Interval::Interval(uint32_t s, uint32_t e) : start(s), end(e) {
+    assert(this->end >= this->start);
+    assert(this->end - this->start <= std::numeric_limits<uint32_t>::max());
+
+    // intervals must be end exclusive so that empty strings can be represented
+    this->length = this->end - this->start;
 }
 
 uint32_t Interval::get_end() const {
-    return start + (uint32_t) length;
+    return start + length;
 }
 
 std::ostream &operator<<(std::ostream &out, Interval const &i) {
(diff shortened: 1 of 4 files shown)


src/interval.cpp
```diff
@@ -7,33 +7,31 @@
...
+    assert(this->end >= this->start);
+    assert(this->end - this->start <= std::numeric_limits<uint32_t>::max());
+
+    // intervals must be end exclusive so that empty strings can be represented
+    this->length = this->end - this->start;
 }
 
 uint32_t Interval::get_end() const {
-    return start + (uint32_t) length;
+    return start + length;
```
mbhall88: Would it not make more sense to just have this function return the `end` property?

rffrancon: Do you mean this: `uint32_t Interval::get_end() const { return this->end; }` ?

mbhall88:
Looking at the sections of the local assembly code where the  interval class is used, I don't think `start` is ever changed. Therefore this should be fine.

src/interval.cpp
mbhall88: yes

mbhall88: I guess, given that we always want `end >= start` we could make `start` and `end` private and have getter/setter functions for them so we can make sure `end >= start` holds.

rmcolq:
The reason for not storing end of an interval was to save RAM - an interval is uniquely defined by start and end, or start and length. Once an interval is created, I don't think it's ever changed. Given how many intervals in the whole thing (they make up paths, which define minikmers, kmernodes, etc) and how high the RAM use is already, I'm reluctant to return to having 3 values in memory unless someone can convince me that it won't increase the RAM.

mbhall88:
We could still just have 2 variables stored, `start` and `end`. Then when we need `length` we can calculate it? I feel like `end` is probably used more frequently?

iqbal-lab:
@rmcolq - current RAM use for mapping cardio to a whole PRG is ~40Gb RAM according to @mbhall88 , which is vastly better than ever before afaict.  Looks good to me, we shoudlk review RAM/speed tradeoff, but I don;t think it's an auto veto on increasing RAM a bit to increase speed. At least, taking 5 days to run is a lot worse than using 50Gb RAM

rmcolq:
True, RAM is better than it was. However don't forget that 40GB was for a single  sample and ultimately we want to compare many.

Also yes we don't want it to take days.



iqbal-lab:
I suggest multi sample can be done sequentially, with one gene at a time in ram

PR #100, opened 2018-11-08
Refactor LocalPRG::shift(.)

rffrancon:
Related to issue #23 .
Please don't merge.

### DIFF
--- a/src/localPRG.cpp
+++ b/src/localPRG.cpp
@@ -68,16 +68,15 @@ std::vector<LocalNodePtr> LocalPRG::nodes_along_path(const Path &p) const {
 
     for (auto it = p.path.begin(); it != p.path.end(); ++it) {
         const auto &interval = *it;
-        uint32_t interval_end = it->get_end();
 
         // find the appropriate node of the prg
         for (auto n = prg.nodes.begin(); n != prg.nodes.end(); ++n) {
-            auto node_ptr = n->second;
-            uint32_t node_start = node_ptr->pos.start;
-            uint32_t node_end = node_ptr->pos.get_end();
+            auto local_node = (LocalNode) *(n->second);
+            uint32_t node_start = local_node.pos.start;
+            uint32_t node_end = local_node.pos.end;
 
-            bool is_overlapping_node = interval_end > node_start and interval.start < node_end;
-            bool is_equal_node = interval.start == node_start and interval_end == node_end;
+            bool is_overlapping_node = interval.end > node_start and interval.start < node_end;
+            bool is_equal_node = interval.start == node_start and interval.end == node_end;
             bool is_null_node = (
                     interval.start == node_start
                     and interval.length == 0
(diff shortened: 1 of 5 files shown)


src/localPRG.cpp
```diff
@@ -247,54 +247,30 @@ std::vector<uint32_t> LocalPRG::build_graph(const Interval &i,
...
             short_paths.pop_front();
 
-            if (n.back()->pos.get_end() == (--(prg.nodes.end()))->second->pos.get_end()) {
-                return_paths.push_back(p);
-            } else if (n.back()->pos.get_end() == p.get_end()) {
-                for (uint32_t j = 0; j != n.back()->outNodes.size(); ++j) {
-                    if (n.back()->outNodes[j]->pos.length == 0) {
-                        short_paths.push_back(p);
-                        short_paths.back().add_end_interval(n.back()->outNodes[j]->pos);
+            if (nodes.back()->pos.get_end() == (--(prg.nodes.end()))->second->pos.get_end()) {
```
rffrancon: @rmcolq What does this conditional mean please?

src/localPRG.cpp
```diff
@@ -304,11 +280,38 @@ std::vector<Path> LocalPRG::shift(Path p) const {
...
         }
     }
 
-    return return_paths;
+    return shifted_paths;
+}
+
+std::vector<Path> LocalPRG::get_extended_paths(const Path &path) const {
+    std::deque<Path> short_paths = {path};
+    std::vector<Path> k_paths = {};
```
rffrancon: @rmcolq What is `k_paths`? Kmer paths?

src/localPRG.cpp
```diff
@@ -247,54 +247,30 @@ std::vector<uint32_t> LocalPRG::build_graph(const Interval &i,
...
-        } else if (p.get_end() != (--(prg.nodes.end()))->second->pos.get_end()) {
-            for (uint32_t i = 0; i != n.back()->outNodes.size(); ++i) {
-                //exp_num_return_seqs += 1;
-                short_paths.push_back(p);
-                short_paths.back().add_end_interval(
-                        Interval(n.back()->outNodes[i]->pos.start, n.back()->outNodes[i]->pos.start));
-            }
-        }
-    }
+    Path truncated_path = path.subpath(1, path.length() - 1);
```
rffrancon: Does this line shift or truncate the path? If truncate, why?

rmcolq: This line takes the truncated path from position 1 along the original path to the end. I want the paths which are shifted along one position in the graph. I have the first `k-1` bases length of this desired path(s) already, so I use them and look for the next `1` base length to add to the end of the path

src/localPRG.cpp
rffrancon: Thanks for the response. What is `k` please?

rmcolq: `k` is an input parameter and determines the length of the (w,k)-minimizers (smallest kmer in window of w consecutive kmers). I usually use `k=15`, but for other kmer-based work values are often e.g. `21`, `31`.

rffrancon: Would `min_kmer_length` be an accurate name?

rmcolq: `k` is `kmer_length`. All kmers have to be exactly `k` bases long. Looking again, `k_paths` are the paths of length exactly `k` and `short_paths` are the paths of length less than `k` which I'm extending to paths of length `k`.

iqbal-lab:
hi @rffrancon !

PR #109, opened 2018-11-08
Re-add changes to local assembly that got lost

mbhall88:
@rmcolq and @rffrancon can you please go through and make sure this is up-to-date with all your changes to.


rmcolq:
I haven't edited those files, so changes all look good to me.

src/denovo_discovery/local_assembly.cpp
```diff
@@ -109,28 +125,17 @@ void get_paths_between_util(const std::string &start_kmer, const std::string &en
...
-        // prevent the number of paths going above a fixed number
-        if (full_paths.size() > g_max_num_paths) {
-            BOOST_LOG_TRIVIAL(debug) << "Too many paths found " << std::to_string(full_paths.size())
-                                     << ". Skipping slice.";
-            full_paths.clear();
-            return;
-        }
     }
 
     for (unsigned int i = 0; i < num_children; ++i) {
```
rffrancon: `for (const auto &child_node: child_nodes) {`
no need for `num_children` variable, right?

src/denovo_discovery/local_assembly.cpp
```diff
@@ -109,28 +125,17 @@ void get_paths_between_util(const std::string &start_kmer, const std::string &en
     auto &child_nodes = tree[start_kmer];
```
rffrancon: perhaps move this alias closer to where it's used?: after the if statement below

rffrancon:
These changes don't touch any code that I've changed from what I can tell.

May I suggest squashing the commits.

mbhall88:
> These changes don't touch any code that I've changed from what I can tell.
> 
> May I suggest squashing the commits.

When I look in my git log there is other commits from you in between some of these commits so I don't see a straightforward way of squashing them - well at least I don't trust myself to do it.

mbhall88:
Ah cool, GitHub gave me the option to squash and merge them.

PR #128, opened 2019-05-13
Some changes to the build

leoisl:
Changed some building flags to make debugging easier;
Added a singularity image to make debugging with gdbserver;

### DIFF
--- a/Singularity_debug.pandora
+++ b/Singularity_debug.pandora
@@ -41,16 +41,6 @@ MirrorURL: http://mirrors.ukfast.co.uk/sites/archive.ubuntu.com/
     cd ..
 
     #============================================
-    # INSTALL PANDORA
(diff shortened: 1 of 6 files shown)


Singularity_debug.pandora
```diff
@@ -41,16 +41,6 @@ MirrorURL: http://mirrors.ukfast.co.uk/sites/archive.ubuntu.com/
     cd ..
```
mbhall88: Change the name of the file to be `Singularity.pandora_debug` otherwise the tags on Singularity Hub will overwrite the other pandora recipe.

rmcolq:
Also we chatted earlier, and thought that we would close this pull request and perhaps start another to merge a subset in. It seems like some files are more suited to the local fork.

PR #134, opened 2019-05-27
#23

leoisl:
Runtime improvement on index #23

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,7 +31,7 @@ bin/*
 *.mod
 *.smod
 
-# Compiled Static libraries
+# Compiled Staticpandora_debug.img libraries
 *.lai
 *.la
 *.a
@@ -67,3 +67,10 @@ scripts/*
 
 # local directory - michael
 vm-build/
+
+# local directory - leandro
+Singularity_debug.pandora
+debug_with_singularity.sh
+pandora_debug.img
+build_debug/
+build/
--- a/.gitignore
+++ b/.gitignore
@@ -74,3 +74,4 @@ build/
 data_issue*
 build_release/
 build_release_with_debug/
+build_profiling/
--- a/.gitignore
+++ b/.gitignore
@@ -74,3 +74,6 @@ debug_with_singularity.sh
 pandora_debug.img
 build_debug/
 build/
+data_issue*
+build_release/
+build_release_with_debug/
--- a/.gitignore
+++ b/.gitignore
@@ -62,9 +62,6 @@ log.txt
 # and files for debugging and sending to other people
 EcoliK12_pandora*
 
-#internal scripts
-scripts/*
-
 # local directory - michael
 vm-build/
 
--- a/.gitignore
+++ b/.gitignore
@@ -31,7 +31,7 @@ bin/*
 *.mod
 *.smod
 
-# Compiled Staticpandora_debug.img libraries
+# Compiled Static libraries
 *.lai
 *.la
 *.a
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "cgranges"]
+	path = cgranges
+	url = https://github.com/lh3/cgranges
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,7 +21,7 @@ else ()
     message(FATAL_ERROR "Compiler ${CMAKE_CXX_COMPILER} has no C++11 support")
 endif ()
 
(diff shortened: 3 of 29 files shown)


leoisl:
> Sounds like it all works based on your testing.

Yeah, but I fear my tests might not be enough, that is why I've hesitated on doing this PR since 1 week ago... I don't really know what could break or not... I verified that `index` is working, and `map` seems to work (only one test done though...). I am wondering, could we have a `dev stable` and a `dev unstable`? At the same time `dev` could allow for unstable builds, I wonder why the sing recipe points to `dev` instead of `master`. I felt that `dev` must then hold a stable build, and hesitated on pushing into it.

> Only thing I would say is maybe reduce the comments as you continue working on the code base. My view is that if something needs lots of comments maybe renaming functions/variables would be a better approach (_a la_ [clean code](https://www.amazon.co.uk/Clean-Code-Handbook-Software-Craftsmanship/dp/0132350882)).

Oh, I was unaware of the [clean code](https://www.amazon.co.uk/Clean-Code-Handbook-Software-Craftsmanship/dp/0132350882) approach. I was always very verbose when coding, commenting as much as I can, but I agree with you on this. I will take your advice into account when working on the code base.

iqbal-lab:
Stop worrying @leandro_ishi, embrace uncertainty. Life is a moving river

rmcolq:
Problem solved, I've merged it!

rmcolq:
No going back... oh wait, this is github!

leoisl:
> Wow! This is a huge amount of great work!
> A couple of thoughts:
> 
> 1. The scripts to compare indexes to my mind should probably be local to you (name in gitignore) as they are not really part of pandora or tests. Unless Michael/Zam disagree, maybe they could be removed from the merge?

You are right, I will remove them!

> 2. In KmerGraph, you have changed the vector of `sorted_nodes` to a set, which is filled with nodes as the graph is created. I think moving forward maybe `sorted_nodes` could be created when its about to be used e.g. in one of the max_path functions. You already have a couple of todo comments in that region to do with it anyway. I think the reason I didn't populate it before was to avoid having pointers in memory twice.

Indeed, this can be a possibility, I will think on this when working on the `max_path` functions. But also this adds an additional layer on `KmerGraph`. Keeping the nodes always sorted is a little bit slower and consumes more memory. Sort the nodes only when they are needed includes the complexity of always remembering that the nodes must be sorted before using them. This has the risk of breaking the code later if someone uses them without sorting (and we probably will). I usually like a lot to control access to the fields through encapsulation (declare `sorted_nodes` as `private` and `public` methods to access it). With encapsulation, `sorted_nodes` is always accessed with for e.g. `get_sorted_nodes()`, instead of directly through `sorted_nodes`, and we can freely change the implementation as we wish. If we keep the interface the same, other parts that access this field through the interface will not break. The implementation is thus abstracted as the public interface is never changed. A big amount of changes I've made comprised of controlling the access to some variables, so that I can change the implementation as I wanted to test things, but this changed the current interface and triggered a lot of changes...

I guess I deviated a bit of the question, but I am unsure about rewriting some classes just to add encapsulation to the code base. As such, improvements can be done a lot easier, but this would be a 2-week work that would not provide us any performance/memory improvement. What do you think? Maybe leave for an eventual refactoring of the code?

> 3. Agree with Michael that comments are good for getting to know code, and slightly less clean than clearer variable names in the long run.

I agree too!

mbhall88:
> > Sounds like it all works based on your testing.
> 
> Yeah, but I fear my tests might not be enough, that is why I've hesitated on doing this PR since 1 week ago... I don't really know what could break or not... I verified that `index` is working, and `map` seems to work (only one test done though...). I am wondering, could we have a `dev stable` and a `dev unstable`? At the same time `dev` could allow for unstable builds, I wonder why the sing recipe points to `dev` instead of `master`. I felt that `dev` must then hold a stable build, and hesitated on pushing into it.  

`master` is supposed to be stable and `dev` is "unstable". So it's ok if something breaks, that's the point of `dev` - to find those breaks and fix them 😄  . When `dev` has been stable for a bit we push to `master`.

> > Only thing I would say is maybe reduce the comments as you continue working on the code base. My view is that if something needs lots of comments maybe renaming functions/variables would be a better approach (_a la_ [clean code](https://www.amazon.co.uk/Clean-Code-Handbook-Software-Craftsmanship/dp/0132350882)).
> 
> Oh, I was unaware of the [clean code](https://www.amazon.co.uk/Clean-Code-Handbook-Software-Craftsmanship/dp/0132350882) approach. I was always very verbose when coding, commenting as much as I can, but I agree with you on this. I will take your advice into account when working on the code base.

They have the book in the library on campus. I would highly recommend having a look.

PR #143, opened 2019-06-24
Path memoization is now done only on pandora index

leoisl:
Path memoization in `LocalPRG::nodes_along_path` method speeds up `pandora index` a lot at the cost of using more memory. `pandora map` and `pandora compare` are mostly limited by RAM usage than speed though.

This PR disables path memoization in all modules but `pandora index`.

This modification brings the RAM usage of `pandora compare` on the klebs PRG with 151 samples down to 17.1GB:

* Runtime:
  * 5h08min with memoization
  * 5h21min without memoization
* RAM usage:
  * 24.5GB with memoization
  * 17.1 GB without memoization

The path memoization speed-up in `pandora compare` is small (13 minutes faster in this example), while the RAM required for the memoization is ~7.5 GB. So, in `pandora map` and `pandora compare` it seems that it is better to disable path memoization.

Added unit tests to test both memoized and non-memoized versions of this method, and assured that `pandora compare` and `pandora index` are producing the same results as before in a real dataset. No real dataset tests were done with `pandora map` though...

### DIFF
--- a/include/localPRG.h
+++ b/include/localPRG.h
@@ -39,6 +39,8 @@ public:
     //VCF vcf;
     std::vector<uint32_t> num_hits;
 
+    static bool path_memoization_in_nodes_along_path_method; //flag to turn on or off path memoization in nodes_along_path method
+
     LocalPRG(uint32_t id, const std::string &name, const std::string &seq);
 
     // functions used to create LocalGraph from PRG string, and to sketch graph
--- a/include/localPRG.h
+++ b/include/localPRG.h
@@ -39,7 +39,7 @@ public:
     //VCF vcf;
     std::vector<uint32_t> num_hits;
 
-    static bool path_memoization_in_nodes_along_path_method; //flag to turn on or off path memoization in nodes_along_path method
+    static bool do_path_memoization_in_nodes_along_path_method;
 
     LocalPRG(uint32_t id, const std::string &name, const std::string &seq);
 
--- a/src/index_main.cpp
+++ b/src/index_main.cpp
@@ -95,10 +95,14 @@ int pandora_index(int argc, char *argv[]) // the "pandora index" command
         }
     }
 
+    //configure logging
     auto g_log_level{boost::log::trivial::info};
     if (log_level == "debug")
         g_log_level = boost::log::trivial::debug;
     boost::log::core::get()->set_filter(boost::log::trivial::severity >= g_log_level);
+
+    //configure path memoization to be done
+    LocalPRG::path_memoization_in_nodes_along_path_method = true;
 
     // load PRGs from file
     std::vector<std::shared_ptr<LocalPRG>> prgs;
--- a/src/index_main.cpp
+++ b/src/index_main.cpp
@@ -95,14 +95,12 @@ int pandora_index(int argc, char *argv[]) // the "pandora index" command
         }
     }
 
-    //configure logging
     auto g_log_level{boost::log::trivial::info};
     if (log_level == "debug")
         g_log_level = boost::log::trivial::debug;
     boost::log::core::get()->set_filter(boost::log::trivial::severity >= g_log_level);
 
-    //configure path memoization to be done
-    LocalPRG::path_memoization_in_nodes_along_path_method = true;
+    LocalPRG::do_path_memoization_in_nodes_along_path_method = true;
 
     // load PRGs from file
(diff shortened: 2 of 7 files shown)


include/localPRG.h
```diff
@@ -39,6 +39,8 @@ class LocalPRG {
     //VCF vcf;
     std::vector<uint32_t> num_hits;
 
+    static bool path_memoization_in_nodes_along_path_method; //flag to turn on or off path memoization in nodes_along_path method
```
mbhall88: Given this is a `bool` maybe the name should be `do_path_memoization_in_nodes_along_path_method` (would also mean you can remove the comment)

src/index_main.cpp
```diff
@@ -95,11 +95,15 @@ int pandora_index(int argc, char *argv[]) // the "pandora index" command
         }
     }
 
+    //configure logging
```
mbhall88: Unnecessary

src/index_main.cpp
```diff
@@ -95,11 +95,15 @@ int pandora_index(int argc, char *argv[]) // the "pandora index" command
...
     auto g_log_level{boost::log::trivial::info};
     if (log_level == "debug")
         g_log_level = boost::log::trivial::debug;
     boost::log::core::get()->set_filter(boost::log::trivial::severity >= g_log_level);
 
+    //configure path memoization to be done
```
mbhall88: comment not needed

src/localPRG.cpp
```diff
@@ -18,6 +18,9 @@
 #define assert_msg(x) !(std::cerr << "Assertion failed: " << x << std::endl)
 
 
+//memoization is always disabled by default
```
mbhall88: comment not needed

test/localPRG_test.cpp
```diff
@@ -272,6 +275,17 @@ TEST(LocalPRGTest, nodes_along_path) {
     EXPECT_EQ(j, l3.nodes_along_path(p3)[1]->id);
 }
 
+TEST(LocalPRGTest, nodes_along_path_without_memoization) {
+    assert(LocalPRG::path_memoization_in_nodes_along_path_method == false);
```
mbhall88: Use the googletest fixture `EXPECT_FALSE` instead of `assert`

test/localPRG_test.cpp
```diff
@@ -272,6 +275,17 @@ TEST(LocalPRGTest, nodes_along_path) {
...
+    nodes_along_path_core_test();
+}
+
+TEST(LocalPRGTest, nodes_along_path_with_memoization) {
+    LocalPRG::path_memoization_in_nodes_along_path_method=true;
+    nodes_along_path_core_test();
+    LocalPRG::path_memoization_in_nodes_along_path_method=false;
```
mbhall88: This test doesn't seem to test anything? There is no assertion of anything being equal or true etc.

leoisl: All tests are done on the `nodes_along_path_core_test` function (https://github.com/leoisl/pandora/blob/improving_compare/test/localPRG_test.cpp#L163-L276), which is called in this test. It tests a bunch of things (i.e. it was the previous unit test). On `TEST(LocalPRGTest, nodes_along_path_without_memoization)`, we just make sure memoization is turned off and call `nodes_along_path_core_test()` (i.e. we test `LocalPRG::nodes_along_path()` without memoization - exactly what was done before). This test tests `LocalPRG::nodes_along_path()` with memoization, to make sure memoization does not introduce any bugs.

include/localPRG.h
leoisl: Done in new commit

src/index_main.cpp
leoisl: Done in new commit

src/index_main.cpp
leoisl: Done in new commit

src/localPRG.cpp
leoisl: Done in new commit

test/localPRG_test.cpp
leoisl: I guess here we need to disable memoization and test `LocalPRG::nodes_along_path()`.

So I think that this this `assert()` is not correct, `EXPECT_FALSE` won't also forcibly disable memoization. So in the next commit, we just forcibly disable it and test the method.

rmcolq:
Looks fine to me! I'll rebase and merge

Issue #145, opened 2019-06-25
Improve inferred maxpath


rmcolq:
During thesis analysis, discovered that the accuracy of inferred mosaic sequences through the PanRG are less accurate at the start of the sequences. This is because there is lower sensitivity to distinguish between similar floats representing the mean log probability of kmer coverages along path. Should update with a parameter to average instead over a fixed buffer size.

PR #142, opened 2019-06-21
Improving pandora map and compare

leoisl:
Main changes: added multithreading, removed some memory leaks, improved RAM usage and performance of both pandora map and compare.

Minor changes: lots of commenting, and some classes refactoring.

Note: old unit tests are working, but should be reviewed. New tests must be added to cover new code.

PR #146, opened 2019-06-25
Update maxpath in kmergraph

rmcolq:
This is for reference only, showing current state of work to do the following:
1. Issue #145 to average log probs over some fixed number of kmers, not all
2. Breaking KmerGraph and KmerGraphWithCoverage into separate files
3. Add tests corresponding to KmerGraphWithCoverage back in
4. Improve variable names as per comments in https://github.com/rmcolq/pandora/pull/142
5. Where obvious, improve code coverage in KmerGraphWithCoverage at the same time

So far have addressed 2 and 1, and am part way through adding tests for the new code. Still to finish this, improve code coverage and improve variable names.


rmcolq:
Right, so the tests of my new functions pass, and line coverage for KmerGraphWithCoverage is up to 92.3 %. Are we happy for a merge? 
@leoisl I haven't yet added an equals operator to KmerGraphWithCoverage. Hoping you may be able to add one if it's needed...?

src/kmergraphwithcoverage.cpp
```diff
@@ -63,67 +63,62 @@ void KmerGraphWithCoverage::set_covg(uint32_t node_id, uint32_t value, bool stra
...
 
-float KmerGraphWithCoverage::nb_prob(uint32_t j, const uint32_t &sample_id) {
-    auto k = get_covg(j, 0, sample_id) + get_covg(j, 1, sample_id);
-    //cout << "j: " << j << " " << nodes[j]->covg[0] << " + " << nodes[j]->covg[1] << " = "
-    //     << nodes[j]->covg[0] + nodes[j]->covg[1] << " = " << k << endl;
-    //cout << "nb_prob(" << nb_r << "," << nb_p << "," << k << ") = ";
-    float ret = log(pdf(boost::math::negative_binomial(nb_r, nb_p), k));
+float KmerGraphWithCoverage::nbin_prob(uint32_t node_id, const uint32_t &sample_id) {
+    auto k = get_covg(node_id, 0, sample_id) + get_covg(node_id, 1, sample_id);
+    float ret = log(pdf(boost::math::negative_binomial(negative_binomial_parameter_r, negative_binomial_parameter_p), k));
```
mbhall88: Hate to be a pest, but what is `ret`?

mbhall88:
Loving these renames :heart_eyes: . Big increase in code readability - coming from someone who doesn't really understand this part of the code.

rmcolq:
Return value. I'll update in a bit.



leoisl:
Hello,

sorry for the delay! I went through the whole code changes and everything is fine for me! Nice work!

> line coverage for KmerGraphWithCoverage is up to 92.3 %

That is nice, thanks a lot!

> @leoisl I haven't yet added an equals operator to KmerGraphWithCoverage. Hoping you may be able to add one if it's needed...?

Sure !

> Loving these renames heart_eyes . Big increase in code readability - coming from someone who doesn't really understand this part of the code.

Me too, it is really more understandable for me now!

leoisl:
@rmcolq Looking at the `KmerGraphWithCoverage` definition, we declare the copy constructor with the default implementation:

https://github.com/rmcolq/pandora/blob/f7257e4096a7831409869d9681b8979349b8c7f6/include/kmergraphwithcoverage.h#L44

As there is nothing customized we should do to copy one KmerGraphWithCoverage object to the other (just copy each attribute), I think the default copy constructor is fine. Do you agree with this?

rmcolq:
I agree about default constructors - maybe if we have a careful definition of equals operator, we can check with a test?

leoisl:
Oh, please pardon me, ignore my previous message... I don't know why wherever it was written "equal operator" I was thinking on `operator=` and copy ctor, but not on `operator==`... I don't know why, really a misunderstanding from my part, will implement `operator==` tomorrow!

PR #147, opened 2019-06-26
Implementing suggestions from https://github.com/rmcolq/pandora/pull/142

leoisl:
These are small changes, mostly renaming of variables suggested in https://github.com/rmcolq/pandora/pull/142 . Still many renamings to do, also because I am used to using `camelCase` instead of  `snake_case` (and I did that a lot in the first 1 and a half month working in pandora, sorry...). But I guess that with some more PRs, we might get close to good variable namings.


rmcolq:
Name changes look good to me. Looks like you've updated loads of my camelCase variable names too. I have no idea when we switched to snake_case, but consistent is good.

mbhall88:
> Name changes look good to me. Looks like you've updated loads of my camelCase variable names too. I have no idea when we switched to snake_case, but consistent is good.

I think I asked you about this way back when I started working on pandora because at the time I think you had been using both and so you decided we should stick with snake_case

PR #152, opened 2019-06-28
Closes #151

### LINKED ISSUE #151: Use `uint16` for recording coverage information
Ok, let's switch to `uint16` for coverage information then - almost no errors and half RAM usage for recording coverage information. Let's go for `uint8` only if RAM becomes a bottleneck.

_Originally posted by @leoisl in https://github.com/rmcolq/pandora/issues/148#issuecomment-506398232_

leoisl:
Closes https://github.com/rmcolq/pandora/issues/151

Switching to `uint16` for coverage information reduced the RAM usage for `pandora compare` on the full klebs dataset from 17.1 GB to 13.5 GB.

This dataset contains only 151 samples, and the more samples, the larger is the reduction, so this change is important for scalability.

### DIFF
--- a/include/kmergraphwithcoverage.h
+++ b/include/kmergraphwithcoverage.h
@@ -22,7 +22,7 @@ class KmerGraphWithCoverage {
 private:
     //each coverage is represented as a std::vector, each position of the vector representing the coverage of a sample
     //each coverage of a sample is represented by a std::pair, for the forward and reverse reads coverage of that sample
-    std::vector<std::vector<std::pair<uint32_t, uint32_t>>> nodeIndex2SampleCoverage;
+    std::vector<std::vector<std::pair<uint16_t, uint16_t>>> nodeIndex2SampleCoverage;
     uint32_t exp_depth_covg;
     float binomial_parameter_p;
     float negative_binomial_parameter_p;
@@ -48,13 +48,18 @@ public:
     virtual ~KmerGraphWithCoverage() = default;
 
     //getter
+    //TODO: we should return a uint16_t instead of a uint32_t, but I did not want to change the interface
+    //TODO: converting uint16_t to uint32_t is safe anyway
+    //TODO: some parts of the code accumulates under the type of this variable, so it might be dangerous to change to uint16_t wihtout careful analysis
+    //TODO: leaving to uint32_t to be safe
     uint32_t get_covg(uint32_t node_id, bool strand, uint32_t sample_id) const;
+
     uint32_t get_num_reads() const { return num_reads; }
     uint32_t get_total_number_samples() const {return total_number_samples; }
 
     //setters
     void increment_covg(uint32_t node_id, bool strand, uint32_t sample_id);
-    void set_covg(uint32_t node_id, uint32_t value, bool strand, uint32_t sample_id);
+    void set_covg(uint32_t node_id, uint16_t value, bool strand, uint32_t sample_id);
     void set_exp_depth_covg(const uint32_t);
     void set_binomial_parameter_p(const float);
     void set_negative_binomial_parameters(const float &, const float &);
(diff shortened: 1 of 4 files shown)


include/kmergraphwithcoverage.h
```diff
@@ -22,7 +22,7 @@ class KmerGraphWithCoverage {
 private:
     //each coverage is represented as a std::vector, each position of the vector representing the coverage of a sample
     //each coverage of a sample is represented by a std::pair, for the forward and reverse reads coverage of that sample
-    std::vector<std::vector<std::pair<uint32_t, uint32_t>>> nodeIndex2SampleCoverage;
+    std::vector<std::vector<std::pair<uint16_t, uint16_t>>> nodeIndex2SampleCoverage;
```
mbhall88: snake_case

include/kmergraphwithcoverage.h
```diff
@@ -47,23 +47,37 @@ class KmerGraphWithCoverage {
...
+    TODO: leaving return type as uint32_t to be safe for now, need a recheck later
+     */
     uint32_t get_covg(uint32_t node_id, bool strand, uint32_t sample_id) const;
     uint32_t get_num_reads() const { return num_reads; }
     uint32_t get_total_number_samples() const {return total_number_samples; }
 
     //setters
     void increment_covg(uint32_t node_id, bool strand, uint32_t sample_id);
-    void set_covg(uint32_t node_id, uint32_t value, bool strand, uint32_t sample_id);
+    void set_covg(uint32_t node_id, uint16_t value, bool strand, uint32_t sample_id);
```
mbhall88: change `strand` to `is_forward`

include/kmergraphwithcoverage.h
```diff
@@ -47,23 +47,37 @@ class KmerGraphWithCoverage {
...
     void set_exp_depth_covg(const uint32_t);
     void set_binomial_parameter_p(const float);
     void set_negative_binomial_parameters(const float &, const float &);
     void set_thresh (int thresh) { this->thresh = thresh; }
     void set_num_reads(uint32_t num_reads) { this->num_reads = num_reads; }
 
     void zeroCoverages() {
-        for (auto &sampleCoverage: nodeIndex2SampleCoverage)
-            sampleCoverage = std::vector<std::pair<uint32_t, uint32_t>>(total_number_samples);
+        for (auto &sampleCoverage: nodeIndex2SampleCoverage) {
```
mbhall88: snake_case

leoisl:
> As a more general statement. I would like if we could move away from having these enormous tests that test a huge number of different states and conditions. A test should test one single case and it should be fairly easy to see what the test is testing from reading the code.
> The test here that you have added realistically should be broken up into _many_ smaller tests. Additionally, the name of a test should tell the reader what the test does and what the expected outcome is. See [this test suite](https://github.com/rmcolq/pandora/blob/4ce991eee84dcc4439e80d89877cd036ab6904c5/test/denovo_discovery/denovo_discovery_test.cpp#L102-L298) for an example of testing a single function.

Indeed, you are right, it is a better approach. I will break this single test into 6 different tests.

include/kmergraphwithcoverage.h
leoisl: Done in new commit.

include/kmergraphwithcoverage.h
leoisl: Done in new commit.

include/kmergraphwithcoverage.h
leoisl: Done in new commit.

leoisl:
All your suggestions were done in new commit, please check if further improvements should be made.

rmcolq:
I'm happy to merge/rebase this in now

Issue #166, opened 2019-07-11
Release build tests don't all pass


mbhall88:

rmcolq:
Would this still be the case if we removed the EXPECT_DEATH tests which check particular asserts?

leoisl:
Sorry for the long post, many things here as I started reading how to deal with this.

* Release build optimises for speed, so `-DNDEBUG` flag is passed when building, which disable all asserts. We can simply remove this flag, which will make all tests pass in the Release build, but then we would surely lose performance, since the code will then run all the asserts.

* It seems that the best thing to do is to develop in Debug build (which we do now), but test the Release build (we test the Debug build only), since it is ultimately what will be run outside of the development environment (see https://stackoverflow.com/questions/3416665/unit-testing-on-a-build-server-release-or-debug-code). Some compiler optimisations might make the Debug executable and Release executable behave slightly different, which might cause bugs (e.g. https://stackoverflow.com/questions/20484449/unit-test-passes-in-debug-build-but-fails-in-release-build ). **So my vote is on testing the Release executable, not the Debug one, to make sure the code which will be executed by others is fine.**

However, as said, to test the Release build we have to enable all asserts, which make us lose performance. There is a reason why `cmake` disable asserts in Release mode: asserts should not be in the Release executable. We need to revisit the `assert`s in pandora. I agree with this post: https://stackoverflow.com/a/409981/5264075 . Assertions are for conditions that should **never** happen when running the code - it is solely to catch developer errors, not user/data errors, and thus can be safely disabled in Release mode, e.g. asserting that the node coverage is the same as the number of reads in a `pangenome::Node` (this error cannot be caused by the user or the data, but only by programmers misusing the `update_node_info_with_this_read` function):

https://github.com/rmcolq/pandora/blob/ef6efe1d9e8231a45b2b3e33f5acc614389d1c7c/src/pangenome/pangraph.cpp#L69-L73

Exceptions are used for run-time error conditions, which could potentially happen when running the tool due to user/data problems, and must be enabled in Release mode, e.g.: the user passing a window size smaller than the kmer size:
https://github.com/rmcolq/pandora/blob/ef6efe1d9e8231a45b2b3e33f5acc614389d1c7c/src/compare_main.cpp#L269

This assert above also is very serious - in the current Release build, we allow the user to do this without any error message.

[...]

Issue #118, opened 2019-04-01
Slow KmerGraph test


mbhall88:
This test takes ~5.5 seconds to execute. Is there a way we could shorten it and still test the same functionality?
https://github.com/rmcolq/pandora/blob/a1d02878fa85b49ee59445ebc220a45ecd54eb7f/test/kmergraph_test.cpp#L367-L378

rmcolq:
It takes ages because it's based on real cases where there were bugs at some point. At the time I didn't know what the bugs were. The shorter one would probably be good enough.

mbhall88:
So if you know what the bugs are now is it possible to make shorter examples that pinpoint them? Or is the length something that triggers the bug?

rmcolq:
From memory it was mostly bugs elsewhere not caught by other tests at the time. The length was not important but problem was with pruning edges which are surplus (arise due to local way we find the next kmer) which resulted in non connectedness when the wrong edges were pruned. A better test would test cleaning of kmer graph instead.



Sent from my Samsung Galaxy smartphone.


-------- Original message --------

iqbal-lab:
Sorry to be obtuse, but is 5s really a problem? Does it run in a loop or some?

mbhall88:
5s is in the best case. Sometimes over 10s. Personally, I feel like a single unittest shouldn't take this long. If no one else agrees though, obviously the majority rules.   

Not saying this is necessarily a high priority issue - but can be a pain when debugging and running the test suite a lot.

PR #168, opened 2019-07-11
Adding a RELEASE_WITH_ASSERTS build configuration

leoisl:
Closes https://github.com/rmcolq/pandora/issues/166 if you are fine with enabling asserts in Release mode (which I strongly think we should do as of now).
Release binary now has to be built with: `cmake -DCMAKE_BUILD_TYPE=RELEASE_WITH_ASSERTS ..`.
All tests in both Release and Debug modes are passing.

Might also close https://github.com/rmcolq/pandora/issues/118 , since with optimization flags tests run a lot faster:
`1/1 Test #1: pandora_test .....................   Passed    6.37 sec`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,9 @@ cmake_minimum_required(VERSION 2.8.7)
 set(PROJECT_NAME_STR pandora)
 project(${PROJECT_NAME_STR} C CXX)
 
+# add a RELEASE_WITH_ASSERTS build type
+set(CMAKE_CXX_FLAGS_RELEASE_WITH_ASSERTS "-O3")
+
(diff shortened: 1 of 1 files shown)


mbhall88:
Happy to merge this but I don't think this is a great long-term solution. I still think we should try and remove/refactor all the asserts re: https://github.com/rmcolq/pandora/issues/166#issuecomment-510495534

leoisl:
> Happy to merge this but I don't think this is a great long-term solution. I still think we should try and remove/refactor all the asserts re: [#166 (comment)](https://github.com/rmcolq/pandora/issues/166#issuecomment-510495534)

Agreed, created an issue for this long-term solution: https://github.com/rmcolq/pandora/issues/169

Issue #165, opened 2019-07-11
Host image on dockerhub


mbhall88:
Singularity Hub is starting to seem like it won't be a good long-term solution for hosting our container images (it has currently been down all week for instance).  

I think [Docker Hub](https://www.docker.com/products/docker-hub) is probably the most versatile solution. We kill two birds with one stone too as we support Docker and Singularity from one location. Additionally, Docker Hub is much more widely used and therefore is maintained a lot better.

leoisl:
I agree on this. Should we also provide a portable linux binary (e.g. download and execute the binary), like it was done in kallisto (see https://pmelsted.wordpress.com/2015/10/14/building-binaries-for-bioinformatics/ ). I did this for DBGWAS, it worked fine.

rmcolq:
If you are happy to build it, YES

leoisl:
Yep, I have already the scripts, so it would be just copy to pandora repo, and change one thing or another I guess...

PR #170, opened 2019-07-24
Add Dockerfile

mbhall88:
First part of resolving #165 .

Once this PR is merged, @rmcolq will need to [setup automated builds on Docker Hub](https://docs.docker.com/docker-hub/builds/) so the image builds and is stored on there.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -61,5 +61,4 @@ RUN mkdir -p build
 WORKDIR build
 RUN cmake -DCMAKE_BUILD_TYPE="$PANDORA_BUILD" .. && make
 RUN ln -s $(realpath pandora) /usr/local/bin/pandora
-# uncomment the below once tests pass for release mode
 RUN ctest -V
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,64 @@
+# PANDORA
+# Pan-genome inference and genotyping with long noisy or short accurate reads
+
+ARG UBUNTU_VERSION="18.04"
+FROM ubuntu:"$UBUNTU_VERSION"
+
+ENV LC_ALL C.UTF-8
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+# INSTALL PANDORA
+#============================================
+ENV PANDORA_BRANCH dev
+ENV PANDORA_GIT="https://github.com/rmcolq/pandora.git"
+
+RUN git clone -b "$PANDORA_BRANCH" --single-branch  --recursive "$PANDORA_GIT"
+WORKDIR pandora
+RUN mkdir -p build
+WORKDIR build
+RUN cmake -DCMAKE_BUILD_TYPE=Release .. && make
```
leoisl: Please build with `-DCMAKE_BUILD_TYPE=RELEASE_WITH_ASSERTS`

Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+RUN ln -s $(realpath pandora) /usr/local/bin/pandora
+# uncomment the below once tests pass for release mode
+# RUN ctest -V
```
leoisl: Then we can enable tests back. I guess this will be the option people will prefer rather than building from source, so I think it is good that we make sure unit tests are passing on it.

Dockerfile
mbhall88: Done in https://github.com/rmcolq/pandora/pull/170/commits/6886d3a875839a1a213b4fa28b02e9038758d589

Dockerfile
mbhall88: Good point. Resolved in https://github.com/rmcolq/pandora/pull/170/commits/6886d3a875839a1a213b4fa28b02e9038758d589

Issue #174, opened 2019-08-11
Fix segfaults with MacOS


rmcolq:
There are issues with OpenMP compatibility with clang/MacOS compilers. The recommended way to get a compiler with OpenMP support on Mac is to download LLVM from http://releases.llvm.org/download.html. However, this doesn't seem to provide full OpenMP support and the result is that while the code compiles fine, when the tests are run it causes a segfault. Stack from debugger when running tests reads:
```
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x7faadb9b93b8)
  * frame #0: 0x0000000110e8a1cc libomp.dylib`int __kmp_acquire_queuing_lock_timed_template<false>(kmp_queuing_lock*, int) + 396
    frame #1: 0x0000000110e8ef39 libomp.dylib`__kmp_acquire_queuing_lock(kmp_user_lock*, int) + 9
    frame #2: 0x0000000110e5045e libomp.dylib`__kmpc_critical_with_hint + 1262
    frame #3: 0x000000010db11baf pandora_test`infer_localPRG_order_for_reads(prgs=size=1, minimizer_hits=std::__1::shared_ptr<MinimizerHits>::element_type @ 0x00007fa848f471a8 strong=2 weak=1, pangraph=std::__1::shared_ptr<pangenome::Graph>::element_type @ 0x00007fa848e121f8 strong=3 weak=1, max_diff=1, genome_size=0x00007fff52578010, fraction_kmers_required_for_cluster=0x00007fff52576fcc, min_cluster_size=1, expected_number_kmers_in_short_read_sketch=4294967295) at utils.cpp:373
[...]

PR #175, opened 2019-08-11
Resolves issue #174 and disable openmp for mac

rmcolq:
As per the similar CD-hit issue, solution is to disable OpenMP (and multithreading) for MacOS.

- Fix includes removing the -fopenmp linker flag in cmakelists file if the OS is Mac.
- Also avoiding the openmp file locks where necessary.
- Slightly surprisingly, no further handling of the omp process sections seems to be needed for tests to pass (and not segfault), when the -fopenmp is not linked.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,12 +24,14 @@ else ()
     message(FATAL_ERROR "Compiler ${CMAKE_CXX_COMPILER} has no C++11 support")
 endif ()
 
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DBOOST_SYSTEM_NO_DEPRECATED -Wall -Wextra")
+
 if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
     set(MACOSX TRUE)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DBOOST_SYSTEM_NO_DEPRECATED -DNO_OPENMP -Wall -Wextra")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DNO_OPENMP")
 else ()
-    find_package(OPENMP 4.5 REQUIRED)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DBOOST_SYSTEM_NO_DEPRECATED -fopenmp -Wall -Wextra")
+    find_package(OPENMP 4.0 REQUIRED)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fopenmp")
 endif()
 
 #print CMAKE debug and release flags to know what we are exactly doing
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -24,7 +24,13 @@ else ()
     message(FATAL_ERROR "Compiler ${CMAKE_CXX_COMPILER} has no C++11 support")
 endif ()
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DBOOST_SYSTEM_NO_DEPRECATED -fopenmp -Wall -Wextra")
+if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
+    set(MACOSX TRUE)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DBOOST_SYSTEM_NO_DEPRECATED -DNO_OPENMP -Wall -Wextra")
+else ()
+    find_package(OPENMP 4.5 REQUIRED)
```
leoisl: I am wondering if we should downgrade the OpenMP version required from 4.5 to 4.0. The reason is that we require C++11 support, and gcc fully supports C++11 from gcc 4.8.1 (https://gcc.gnu.org/projects/cxx-status.html#cxx11), but OpenMP 4.5 is only supported in gcc 6.1 (https://www.openmp.org/resources/openmp-compilers-tools/). Downgrading to OpenMP 4.0 means that the user needs only gcc 4.9.0 to compile. Although I did not test if it works with OpenMP 4.0, I am guessing so since there are no features exclusively from OpenMP 4.5 that we use (see https://www.openmp.org/wp-content/uploads/openmp-4.5.pdf pages 348-350).

CMakeLists.txt
```diff
@@ -24,7 +24,13 @@ else ()
     message(FATAL_ERROR "Compiler ${CMAKE_CXX_COMPILER} has no C++11 support")
 endif ()
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DBOOST_SYSTEM_NO_DEPRECATED -fopenmp -Wall -Wextra")
+if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
+    set(MACOSX TRUE)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DBOOST_SYSTEM_NO_DEPRECATED -DNO_OPENMP -Wall -Wextra")
```
leoisl: Minor thing (no need to do this if you feel there is no need to): move all non-OpenMP-related flags before the `if` (those that will be included regardless if the compiling machine is Linux or Mac), and in the `if` just add `-fopenmp` or `-DNO_OPENMP`. This is to avoid us to have to consistently change both lines when compiling flags change.

CMakeLists.txt
rmcolq: I've downgraded to 4.0. I can't test if it compiles fine on my Mac, so if it doesn't we can change the number again later.

CMakeLists.txt
rmcolq: Agreed. Done.

leoisl:
All fine for merge for me

PR #184, opened 2019-10-11
Closes #183 - Fixes bug in LocalPRG::append_kmer_covgs_in_range

### LINKED ISSUE #183: Fix bug in LocalPRG::append_kmer_covgs_in_range
Just to have an entry in `pandora` repo and a short description:

The bug lies in https://github.com/rmcolq/pandora/blob/d2dd2286d7a657095b73104dd11d8b7655509537/src/localPRG.cpp#L1187-L1190 when computing how many bases there are in the `local_path` before we reach the first (non-trivial) kmer in the `kmer_path` (https://github.com/rmcolq/pandora/blob/dev/src/localPRG.cpp#L1197-L1210). In the H131800734, the first (non-trivial) local node had interval `[3, 8]`, and the first kmer in `kmer_path` had interval `[0, 29]`. This would trigger an underflow in the `added` variable (https://github.com/rmcolq/pandora/blob/d2dd2286d7a657095b73104dd11d8b7655509537/src/localPRG.cpp#L1207), making it an enormous value, and thus no coverage would be added.

leoisl:
This PR closes #183 and https://github.com/iqbal-lab/pandora1_paper/issues/101 . We still did not manage to generate the new ROC curves due to cluster issues to check the impact of this fix, but we think this will eventually need to be merged anyway.

I had put the three of you as reviewers because this PR is not just a bugfix, it also describes how I will refactor / make unit tests from now on (i.e. it will impact in `pandora` evolution). Please pay attention to the refactoring of the `LocalPRG::append_kmer_covgs_in_range` method, and the new tests related to `LocalPRG::get_number_of_bases_that_are_exclusively_in_the_previous_kmer_node` and `ocalPRG::get_number_of_bases_that_are_exclusively_in_the_previous_kmer_node`. I will be refactoring `pandora` slowly whenever a bugfix is needed or taking one day/week to do some refactor. Refactoring has also some fraction of personal coding preference, and it also means investing time to improve code quality and maintenance, but no new features/performance improvement is really done.

In the specific case of this PR, the bugfix could have been done by changing only one condition, and it would take 5 minutes and one line change. Refactoring the code and adding unit tests for each new method added during refactoring extended this to ~3-4 hours. More unit tests means that the code is more robust, and this is nice, but the refactoring is more subjective to judge.

Tell me how should I proceed from here: should I spend time for refactoring + more unit tests like done in this PR, or should I focus solely on solving bugs as fast as possible, and investing time in new features/performance improvements? Also, is there a ratio of time spent on refactoring and on evolution (new features/performance improvement) that you are happy with?

Please don't hesitate on being very critical to any of the additions in this PR, it will actually help me and it will increase the code quality in the next PRs.

### DIFF
--- a/src/localPRG.cpp
+++ b/src/localPRG.cpp
@@ -5,6 +5,7 @@
 #include <algorithm>
 #include <numeric>
 #include <cstdlib>
+#include <utility>
 
 #include <boost/log/trivial.hpp>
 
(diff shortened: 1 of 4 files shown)


src/localPRG.cpp
```diff
@@ -1183,62 +1184,110 @@ LocalPRG::find_alt_path(const std::vector<LocalNodePtr> &ref_path, const uint32_
...
+}
+
+
+std::pair<std::vector<uint32_t>, std::vector<uint32_t>>
+LocalPRG::append_kmer_covgs_in_range(const KmerGraphWithCoverage &kmer_graph_with_coverage,
+                                     const std::vector<KmerNodePtr> &kmer_path,
+                                     const std::vector<LocalNodePtr> &local_path, const uint32_t &range_pos_start,
+                                     const uint32_t &range_pos_end, const uint32_t &sample_id) const {
+    assert(kmer_path.size() >
+           1); // this is an assert because it is the programmers responsibility to ensure that the kmer_path given to this function has at least size 1
```
mbhall88: I think formatting moved the 1 to a new line.

src/localPRG.cpp
```diff
@@ -1183,62 +1184,110 @@ LocalPRG::find_alt_path(const std::vector<LocalNodePtr> &ref_path, const uint32_
...
+        number_of_bases_that_are_exclusively_in_the_previous_kmer_node +=
+                current_kmer_node->path.get_start() - previous_kmer_node_path_interval_iterator->start;
+    }
+
+    return number_of_bases_that_are_exclusively_in_the_previous_kmer_node;
+}
+
+
+std::pair<std::vector<uint32_t>, std::vector<uint32_t>>
+LocalPRG::append_kmer_covgs_in_range(const KmerGraphWithCoverage &kmer_graph_with_coverage,
```
mbhall88: The name of this function is misleading now. It says append kmer coverages, but it *returns* coverages. Better name might be `forward_and_reverse_kmer_covgs_in_range`

test/localPRG_test.cpp
```diff
@@ -1316,6 +1316,303 @@ TEST(LocalPRGTest, find_alt_path) {
 
 }
 
+// TODO: refactor some tests here to make use of mocks
```
mbhall88: Rather than have the TODO here, be more specific about which tests need refactoring and tag them with the TODO

src/localPRG.cpp
leoisl: Yeah, I tried moving the comments out to check if it would format the 1 back to the previous line, but no success

src/localPRG.cpp
leoisl: Oh, that is actually an important catch! Refactoring it

test/localPRG_test.cpp
leoisl: I should have removed this TODO, as unit tests might be refactored to be compliant with the conventions, if we accept to follow them. So all unit tests will be refactored, even if the majority will be minor refactoring (e.g. declaring common setups in Fixtures, maybe using mock, and splitting a complex unit test into several simple ones).

Issue #158, opened 2019-07-08
Setup continuous integration


mbhall88:
I think we should setup continuous integration so that with all the PRs etc. we have going on it will be easy to see whether tests do indeed pass for all changes. With the added benefit that the testing is done in a clean environment. I think for our current case our best free options are the following:
1. [Travis CI](https://travis-ci.org/) - the free plan has a 120 minute limit for a single build, but unlimited build minutes across builds. (There's also some perks for students).
2. Migrate the code base to GitLab. GitLab has integrated CI plus a huge bonus that it also has a container registry - meaning we can have the singularity image built within GitLab also. They provide [free access to Ultimate and Gold features for open-source projects and educational institutes](https://about.gitlab.com/2018/06/05/gitlab-ultimate-and-gold-free-for-education-and-open-source/). Additionally, the [GitLab integration tool](https://docs.gitlab.com/ee/user/project/import/github.html#using-the-github-integration) will bring all issues, PRs, wikis, labels etc. across as well.

Please add any other options that I have missed or you feel would be better. Also, if you feel we don't need CI then you can also voice that too.

iqbal-lab:
i guess if it has the features it will win, the gitlab ui is ugly

rmcolq:
We already have Travis CI toggles in the README for master and dev. Although I have a feeling that in some cases the build can fail and it still works

rmcolq:
Wow, the last time Travis actually calculated something for the build was 3 years ago!

mbhall88:
> i guess if it has the features it will win, the gitlab ui is ugly

Interesting, I actually prefer the GitLab UI to the GitHub one.  

So are you saying Travis or GitLab @iqbal-lab ? Couldn't tell from your response.

leoisl:
Continuous Integration really seems to be essential now! In my opinion, we should really do this, but whether we should stay on Github with Travis CI or migrate to GitLab, I don't have any strong preference... Thus I leave this decision mostly to you.

rmcolq:
Personally I'd favour keeping the code where it is an fixing up Travis CI.

iqbal-lab:
That is my preference too, but i am aware i don't know gitlab well. can we discuss when we next meet. i see the container is a win

mbhall88:
Another option to add into the list (although it is in beta) is https://github.com/features/actions

PR #189, opened 2019-10-18
Reformat project with `clang-format`

### LINKED ISSUE #188: Reformat project with ClangFormat
Continuing discussion from https://github.com/rmcolq/pandora/pull/186#pullrequestreview-303213542

> The PR is actually small (if I did not miss anything). What makes it very large is the Clang formatting (I guess >90% of the lines are formatting changes, not code logic changes). It made reviewing complicated since sometimes I would skip some parts of the review by rapidly scanning the changes which seemed like only Clang formatting changes.

> I guess we acknowledged this after making these commits (some months ago) and decided to make a big PR with only Clang formatting changes, and then proceed from there. I am willing to do this as soon as possible so that it is easier to review future code, but we still have some WIP branches in pandora. If we apply the Clang formatting in the whole code base now, I guess we will get a bunch of conflicts when merging these branches, which might be hard to solve. So we are thinking on applying Clang Formatting when the code is in a stable state (i.e. probably when finishing the paper).

> **What do you think about this? When should we apply Clang formatting?**

So team, when are we going to do this? I think we have dumped a lot of issues in the "waiting-until-the-code-is-stable" basket. I vote for biting the bullet and doing it.

mbhall88:
Reformatted all files with `clang-format` [closes #188 ].

Additionally, added a format check step to the Travis CI build (and tweaked some other things).

I have also set up a pre-commit git hook locally. What this does it run a check before you can commit locally to see if you have formatted all of the files associated with the commit.

Add the following code snippet inside the pandora project directory as `.git/hooks/pre-commit`

```sh
#!/bin/sh
# pre-commit hook that checks the files being commited have been formatted with 
# clang-format and the project style
git diff --cached --name-only --diff-filter=ACMRT |
  grep -E '\.cpp|\.h' |
  xargs -n1 clang-format -style=file -output-replacements-xml |
  grep "<replacement " >/dev/null
if [ $? -ne 1 ]; then
    echo "PreCommitError: The .cpp and .h files you are trying to commit did not match clang-format required style."
    exit 1
fi
```
and then run `chmod +x .git/hooks/pre-commit`

Also, if you haven't set it up already, you can tell CLion to use `clang-format as it's engine for reformatting https://www.jetbrains.com/help/clion/clangformat-as-alternative-formatter.html

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,12 +10,23 @@ include:
         packages:
           - clang-format-8
 
-before_install:
-  - find src/ include/ test/ -type f \( -iname \*.h -o -iname \*.cpp \) | xargs -I _ clang-format -style=file -output-replacements-xml _ | grep -c "<replacement " >/dev/null
-  - if [ $? -ne 1 ]; then echo "Not all source and header files are formatted with clang-format"; exit 1; fi
+jobs:
+  include:
+    - stage: "Lint"                # naming the Tests stage
+      name: "Format"            # names the first Tests stage job
+      script:
+        - find src/ include/ test/ -type f \( -iname \*.h -o -iname \*.cpp \) | xargs -I _ clang-format -style=file -output-replacements-xml _ | grep -c "<replacement " >/dev/null
+        - if [ $? -ne 1 ]; then echo "Not all source and header files are formatted with clang-format"; exit 1; fi
+    - stage: "Build"
+      script: echo "This is where we build Pandora"
+    - stage: "Test"
+      name: "Unit tests"
+      script: echo "This is where we run the unit tests"
 
-script:
-  - bash install.sh
+stages:
+  - "Lint"
+  - "Build"
+  - "Test"
 
 notifications:
   email: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,22 @@
 language: cpp
-compiler: g++
+compiler: gcc
+os: linux
+dist: bionic
+
+include:
+  - os: linux
+    addons:
+      apt:
+        packages:
+          - clang-format-8
 
 before_install:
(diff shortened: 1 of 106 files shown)


mbhall88:
Tagged as WIP as the formatting has messed up some of the namespaces I think and now it won't compile... Will try and fix quickly.

mbhall88:
Working towards updating the CI [#158 ] I have set up three stages for the CI to pass: Lint, Build, and Test.  

As this PR is just concerned with formatting I have only put real commands in the Lint stage in a job called Format. If we also want to add a linting job within the stage (using `clang-tidy`) we can do that later too. For now, the Build and Test stages only have `echo` statements. So, although they pass, they don't do anything at this stage. That is work for a separate PR.

With this set up Lint gets run first, so if the code being pushed hasn't been formatted to our specified style in `.clang-format` then the build will fail.  

@rmcolq I have enabled all of this on my Travis fork, but I think you might need to also update your Travis settings for pandora so that it runs Travis on PRs etc.

rmcolq:
When you say you have Travis working locally, do you still have an `install.sh` file? Looks like I need to make one again for this to run on PR

mbhall88:
> When you say you have Travis working locally, do you still have an `install.sh` file? Looks like I need to make one again for this to run on PR

No, there isn't an `install.sh` anymore. You wont be able to get it to run on this PR as the travis config doesn't exist on your branch. Also, as `dev` will be quite different from `master` I'm not sure if maybe we need to specify something about branches in the travis settings. Let me take a look.

mbhall88:
So by default Travis should run on all branches provided that option is selected in the Travis Web interface settings (normally on by default). If we only want to build on specific branches we can [specify that in the config] (https://docs.travis-ci.com/user/customizing-the-build/#safelisting-or-blocklisting-branches).

Issue #191, opened 2019-10-30
Fix singularity recipe


leoisl:
Logging this to be done in the next PR.

Singularity recipe should build `pandora` with `RELEASE_WITH_ASSERTS` build type and we should put the tests back.

mbhall88:
Is this not superfluous given we now have a `Dockerfile`?

leoisl:
Oh, it indeed is, but then we should remove the Singularity recipe or keep it consistent with the `Dockerfile`. Should we push the docker image to dockerhub and update the way we pull the image from singularityhub to dockerhub? The image from singularityhub does not include asserts right now, which is essential (otherwise we might get very weird bugs in case of runtime errors, since they are checked with asserts and not exceptions).

mbhall88:
Well in #170 I said 

> Once this PR is merged, @rmcolq will need to setup automated builds on Docker Hub so the image builds and is stored on there.

Then @rmcolq closed #165 . This would suggest it has been pushed to Docker Hub, but I haven't looked.

rmcolq:
It has been pushed to Dockerhub (I had setup automated builds but it went wrong. It does seem to be building dev now). Are you suggesting having a singularity recipe file which just pulls the dockerhub image?

leoisl:
Oh sorry, I got confused by seeing the `Singularity.pandora` file and I forgot about the `Dockerfile`. I guess we should remove the `Singularity.pandora` file and put in the readme that it is hosted in dockerhub instead of singularityhub?

rmcolq:
Ok, will do.

PR #196, opened 2019-11-18
Update singularity docker

rmcolq:
Addresses issue #191 about fixing singularity image (removes it and replaces with docker image)

Looks like an old commit has been added to this branch too. I will remove it.


rmcolq:
Actually it looks like i have to make a new commit and reverse it. Do we want the extended information (it was added in another branch for debugging) or shall I go ahead and put back to normal?

leoisl:
I am fine with keeping the coverage info to the header of consensus fastq, this will be relevant to the identification of FP genes

leoisl:
Oops, you just undid the commit, it is fine also, we can merge it later! and this PR is better now!

rmcolq:
Ok to merge?

leoisl:
seems fine for me!

PR #199, opened 2019-11-20
fix compile error on mac (cstring vs string header) and modify matrix…

rmcolq:
… back to have logical format


leoisl:
If you happen to have other Mac-related compilation/linking/running issues, and don't really want to bother fixing them, please consider development in a containerized environment. In https://github.com/leoisl/pandora_dev_support/tree/master/containerized_dev you can find for either `docker` or `singularity`. I highly recommend `docker`, as it is fully integrated with CLion. It is almost as developing locally. And this is not a VM, it is a container, which makes things a lot faster and do not need preallocation of resources, and `docker` runs fine on Mac. You should get exactly the same dev environment as mine if you use this (I do not even develop on my local environment although I am on Linux, because the containerization is the only way in C++ to isolate your dev environment).

rmcolq:
Awesome, will do next time I get a compilation/linking issues!

mbhall88:
Is there any reason this hasn't been merged?

rmcolq:
@leoisl wanted to wait until the major one was done

PR #211, opened 2020-03-12
Refactor vcf + regenotype only records in maximum likelihood path

leoisl:
Sorry for putting two changes in a single PR.

This PR includes the VCF refactoring and adds a flag that allows the regenotyping to be only on the maximum likelihood path.

The changes are too large to be reviewable. This is a shame, since we lose Code Reviewing, which can be crucial for PRs. I actually do not know how to effectively do code reviews with very large changes... One thing that might assure us that this code is fine to merge is that we have more tests now (751 vs ~600) than we had before, and this is the code we have been debugging in our 4way analysis since late November.

I would merge into master after the 4 or n-way shows it is fine though...


.clang-format
```diff
@@ -1,4 +1,5 @@
 ﻿---
 BasedOnStyle: WebKit
 ColumnLimit: 88
+SortIncludes: false
```
mbhall88: I don't like adding this.

leoisl: Yeah, me neither. Letting `clang-format` sort includes incurs into compilation issues. Your fix solve them partially, but there are more other issues to solve. I think solving these issues manually is not an option (we don't want to do this everytime we clang-format the code base). So I think we have two options:
1. Reject the PR until we can correctly solve the include dependencies;
2. Accept PR and push the solution of include dependencies for later;

rmcolq: I vote for 1 purely so we don't end up putting off the merge indefinitely again...

mbhall88: I vote for 2. 

But I don't understand how `clang-format` is to blame? If our `include` order is causing compilation issues then we must have some messed up dependencies in our files and that should be fixed.  

I will raise an issue for this.

PR #220, opened 2020-07-31
Global and local genotyping

leoisl:
`--genotype` flag has now two modes: global (ML path oriented) or local (coverage oriented).


mbhall88:
@leoisl are you ok to merge this?

leoisl:
Yeah, this is the version we have been evaluating, and that will be in the paper.

PR #223, opened 2020-08-24
Change fastx parser to klib/kseq

### LINKED ISSUE #198: Fastaq header parsing incorrect
https://github.com/rmcolq/pandora/blob/7609f5ffdc3cc57f590fca1e2a4808a11be5201d/src/fastaq_handler.cpp#L48-L53

If my header is `>10001-12000 max_nest=5 min_match=7` then pandora says the sequence name is `10001-12000 max_nest=5 min_match=7` when it should be `10001-12000` (title) and the rest is considered description - see https://academic.oup.com/nar/article/38/6/1767/3112533.

mbhall88:
We do not handle fastx files very robustly and allow invalid files. Additionally, as mentioned in #198, we don't correctly parse the sequence identifier. This PR aims to switch fastq handling to (predominantly) [`klib`](https://github.com/attractivechaos/klib). As much as possible, I have tried to keep API changes to `FastaqHandler` as minimal as possible.

**Note: These changes may be incompatible with indices built with previous versions.** The reason for this is that if the PRG that was indexed was

```
>prg1 min_match=7 max_nest=5
ACGT
```

then the name of the kmer PRG for that will be `kmer_prgs/01/prg1\ min_match=7\ max_nest=5.k15.w14.gfa`.  
With the changes in this PRG, the file will now be `kmer_prgs/01/prg1.k15.w14.gfa`.

## Added

- `is_closed` method on `FastaqHandler`
- `kseq.h` from [`klib`](https://github.com/attractivechaos/klib)

## Changed

- Some of our test fastq files are invalid (i.e. truncated quality strings etc.) as we don't pay attention to quality strings when parsing. As klib does, I have changed it so that these files have quality strings the same length as their corresponding sequence.
- renamed `get_id` to `get_nth_read` as the original name confuses terminology (I normally consider id to be sequence identifier - which this function takes no note of).
- `get_next` now throws `std::out_of_range` if it is called after all reads have been read. The convention now is to use this function call in a `try-catch` block. Code has been refactored so all current usage of this function is in one of these blocks.
- When `get_nth_read` receives an index greater than the number of reads in the file, it will throw an [`std::out_of_range`](https://en.cppreference.com/w/cpp/error/out_of_range) exception now. This is to keep consistent with the new functionality of `get_next`

## Removed

- `skip_next` method on `FastaqHandler`. This method was only being used by `get_id` and all it is doing is calling `get_next`. It seemed a bit useless and removing it didn't cause any tests to fail...
- two `print` methods in `fastaq_handler.cpp` that were not being used anywhere
- some log messages that don't seem useful

closes #198


mbhall88:
I am currently running the tip of this branch on my TB data. Some samples are up to generating de novo paths so I guess the changes haven't caused any major errors. Might be nice if you want to test it out on a sample you have some "expected" output for? I have a docker container built for the tip at `docker://mbhall88/pandora:04beed5`

src/fastaq_handler.cpp
```diff
@@ -1,176 +1,77 @@
...
 }
 
 void FastaqHandler::close()
 {
-    BOOST_LOG_TRIVIAL(debug) << "Close fastaq file";
-    fastaq_file.close();
+    if (!this->is_closed()) {
+        this->closed_status = gzclose(this->fastaq_file);
+    }
 }
```
leoisl, line 75: I might be being too pedantic here, if so, please ignore this comment.

I would prefer to change from `int closed_status` to `bool closed`. `int closed_status` implies that to understand the code, we have to know the return values of `gzclose()`. I also don't understand why `closed_status` is initialised with default 3. I tried to check what 3 means as a `closed_status` in the docs: https://refspecs.linuxbase.org/LSB_3.0.0/LSB-Core-generic/LSB-Core-generic/zlib-gzclose-1.html but I did not find any documentation in it. Also, we could have errors when closing this file, and we are not dealing with it (e.g. `gzclose()` can return `Z_STREAM_ERROR`, `Z_ERRNO` or `Z_BUF_ERROR`), and we would just say it is not closed.

I think it would be simpler and clearer for us in the future to maintain this code if `int closed_status` becomes `bool closed`, initialised with `false` by default. When closing the file, if the close was successful (i.e. `gzclose()` returned `Z_OK`), then `closed` becomes `true`. If `gzclose()` was unsuccessful, then an exception is thrown.

include/fastaq_handler.h
```diff
@@ -4,19 +4,27 @@
...
 
 struct FastaqHandler {
-    bool gzipped;
-    std::ifstream fastaq_file;
-    boost::iostreams::filtering_istreambuf inbuf;
-    std::istream instream;
-    std::string line;
+private:
+    int closed_status; // see
+                       // https://github.com/attractivechaos/klib/blob/928581a78413bed4efa956731b35b18a638f20f3/kseq.h#L171
```
leoisl, line 22: Is this documentation correct? I think `closed_status` refers to return values of `gzclose()`, not return values of `kseq_read()`

include/fastaq_handler.h
```diff
@@ -4,19 +4,27 @@
...
+    kseq_t* inbuf;
+
+public:
+    const std::string& filepath;
```
leoisl, line 26: Please change to `const std::string filepath;`. I will explain why there is a bug if we leave `const std::string& filepath;`, although it is not a serious bug as we basically don't use the `filepath` variable later (only if there is an error on reading a read from the file... but if in the future we want, for example, to reopen the file, and try to use this variable, we will have bugs). Apologies if this is no novelty to you!

I ran this small test on my machine, that basically creates a `FastaqHandler fh` and prints `fh.filepath` to the screen:
```
TEST(FastaqHandlerTest, filepath_test)
{
    FastaqHandler fh(TEST_CASE_DIR + "reads.fq.gz");
    std::cout << "filepath: " << fh.filepath << std::endl;
}
```

This is the output:
`filepath: �^K��U  st/test_cases/reads.fq.gz`

but should have been:
`filepath: ../../test/test_cases/reads.fq.gz`

There are some corrupted bytes at the start of the string. The bottom line is that we are referring to a memory that is not allocated anymore (i.e. `filepath` refers to a `string` that does not exist anymore), so the process can and might overwrite on that memory, because it sees it as available. This might not be reproduceable in your end, it might output the correct string, or a totally different string, it depends on the execution.

This is why this happens. On this line:
`FastaqHandler fh(TEST_CASE_DIR + "reads.fq.gz");`

c++ creates a _temporary, unnamed_ `std::string` with value `../../test/test_cases/reads.fq.gz` and gives it to the `FastaqHandler` constructor. To simplify, let's call this `string` by `s` (although it is an unnamed variable) and let's say it is at memory address `m`. In the constructor, we assign `filepath <- s`. As `filepath` is a `const std::string&`, we are telling c++ that the `std::string` at memory address `m` now has a new name: `fh.filepath` (`&` in c++ roughly means make alias). So, `fh.filepath` now refers to a `std::string` at memory address `m`. After `FastaqHandler` constructor ends execution, and before the next instruction is executed, c++ sees that the `string` it created at address `m` is just a _temporary, unnamed_  object created just to call the constructor. As we haven't named this variable in the test code (in the context where we created it), its life cycle is just this one line. Thus, c++ cleans up this variable, by calling the destructor on this string and releasing its memory. As the memory occupied by the string is now free, further memory allocations might overwrite the values at memory address `m`. The issue is that the `fh` object is still alive, and `fh.filepath` refers to a `string` that already got destroyed, and thus the bug.

By removing `&` from the variable definition, we tell c++ to make a variable copy, instead of a variable alias. Then, if the original copy gets destroyed later, we don't care, as we have our own local copy.

Not a big issue anyway, as we just use `filepath` in an error logging message, excluding the constructor code, but could become an issue in the future. Hope this was somehow useful... c++ in annoyingly error prone, I think all these issues is avoided in rust with the borrow checker... I am not sure, as I don't know rust, but I am seriously thinking about the switch, because bugs like this always slip through, because the language allows it, and they are hard to find and debug...

src/fastaq_handler.cpp
```diff
@@ -1,176 +1,77 @@
...
-        read.push_back(file);
-    }
+    int read_status = kseq_read(this->inbuf);
 
-    // Print read vector
-    for (i = 0; i < read.size(); i++) {
-        std::cout << read[i];
+    bool no_more_reads_available = read_status == -1;
+    if (no_more_reads_available) {
+        return;
```
leoisl, line 30: What do you think about throwing a `std::out_of_bounds` exception if the user try to read sequences when there are no more sequences? Here we don't do anything, i.e. `this->name` and `this->read` refer to the last sequence read, it seems to me a `std::out_of_bounds` would be better? I am not sure, what do you think?

src/fastaq_handler.cpp
```diff
@@ -1,176 +1,77 @@
...
         }
     }
 }
 
 void FastaqHandler::close()
 {
-    BOOST_LOG_TRIVIAL(debug) << "Close fastaq file";
-    fastaq_file.close();
+    if (!this->is_closed()) {
+        this->closed_status = gzclose(this->fastaq_file);
```
leoisl, line 73: I think we should also call `kseq_destroy(inbuf)` to avoid memory leaks

leoisl (changes requested):
Thanks a lot for this, nice PR! I made some comments, feel totally free to accept or reject the proposed changes.

I now prefer this solution to fasta/q parsing than including Seqan. It does the job very well, and it is a lot more lightweight, and the tooling overhead is almost 0, due to `kseq` being a single-file header-only library.

leoisl:
> I am currently running the tip of this branch on my TB data. Some samples have finished so I guess the changes haven't caused any major errors. Might be nice if you want to test it out on a sample you have some "expected" output for? Don't worry if not, I can run the current tip of `dev` on one of the samples tomorrow and see if the output is "the same" as that from this new klib version. I have a docker container built for the tip of this PR at `docker://mbhall88/pandora:04beed5`

It is fine to push unstable versions to dev branch, and the version we are using in the paper is properly tagged. Your test suite is also comprehensive. And if there is any outstanding issue with fasta/q parsing, your real data test should catch it... If you get the same results as in the tip of `dev`, looking at the code and the test suite, I personally feel like the implementation is correct, and should be merged

src/fastaq_handler.cpp
mbhall88: You're right. I will rework this! I love code review!

include/fastaq_handler.h
mbhall88: Yeah, not sure what happened here... :confused:

include/fastaq_handler.h
mbhall88: Wow, that is an awesome explanation! I did not know that. I also have no idea why `filepath` was a reference anyway.

src/fastaq_handler.cpp
mbhall88: Hmm, this is a good point. Probably a good idea to throw an exception. It scares me to think we could be thinking we are getting more reads when we are actually just reusing the last read multiple times (I think we have tests that cover this, but throwing an exception is probably best).

This raises another thing that has been annoying me. I feels like `get_next` would be best to return some `Record` class for the read, rather than invisibly altering internal member variables. However, this is a much larger change...

leoisl: > It scares me to think we could be thinking we are getting more reads when we are actually just reusing the last read multiple times (I think we have tests that cover this, but throwing an exception is probably best).

Yeah, it also now scares me... the "convention" of `get_next()` functions is to throw errors or return a special value when there is nothing else to get, and we might get into trouble in the future if we don't remember that this `get_next()` don't do this...

> This raises another thing that has been annoying me. I feels like get_next would be best to return some Record class for the read, rather than invisibly altering internal member variables. However, this is a much larger change...

Yeah, I also agree... Maybe not a Record class, that does not exist in `pandora`, so we would need to create and refactor everywhere where `get_next()` is called, but maybe just a tuple with the information about what was read, or a `Seq` object (https://github.com/rmcolq/pandora/blob/master/include/seq.h).

PS: returning a `Seq` object won't work because the FastaqHandler handles more than fasta or fastq files. This might be sth important we might have missed, but it seems things did not break, so all is actually fine.
What I mean is that FastqHandler objects are used to handle PRG files: https://github.com/rmcolq/pandora/blob/6f6af4b7a137a158a97806e64c58ad8febb9ccc3/src/utils.cpp#L116-L138
... and PRG files are fasta-like files, but PRG sequences != DNA. I think some libraries might break if we try to read PRG files as fasta files (because of the numbers, spaces, etc...). Good thing that `kseq` does not seem to break, and handle PRG sequences just fine... I think SeqAn would break.
We definitely can't return a `Seq` object because when a `Seq` is created, `pandora` wants to automatically sketch it, which does not work for PRG sequences...

Bottom line is that I also agree with you on this point, and could be improved, but I am not sure if we should change anything now...

leoisl: PS: I agree the Record class to concretely represent a Record in a Fasta/q file is the best solution, but also the one needing most work... was trying to find some middle-ground solutions... I guess whatever you decide to do is fine by me!

mbhall88:
I have implemented the great suggestions @leoisl and updated the PR "changelog" accordingly. Let me know if you seen any issues with my changes.

src/fastaq_handler.cpp
```diff
@@ -1,176 +1,83 @@
...
-    while (num_reads_parsed <= id) {
-        get_next();
-        if (eof()) {
-            break;
+    while (this->num_reads_parsed < one_based_idx) {
+        try {
+            this->get_next();
+        } catch (std::out_of_range& err) {
+            throw std::out_of_range("Requested a read past the end of the file.");
         }
```
leoisl, line 63: Very minor: I think there is no need to enclose this `get_next()` in a `try` `catch` block as it is catching an exception, rebuilding and rethrowing it

src/fastaq_handler.cpp
```diff
@@ -1,176 +1,83 @@
...
-            inbuf.push(boost::iostreams::gzip_decompressor());
-            gzipped = true;
-        }
-        inbuf.push(fastaq_file);
-    } catch (const boost::iostreams::gzip_error& e) {
-        std::cerr << "Problem transfering file contents to boost stream: " << e.what()
-                  << '\n';
+    this->fastaq_file = gzopen(filepath.c_str(), "r");
+    if (this->fastaq_file == nullptr) {
+        throw "Unable to open " + this->filepath;
```
leoisl, line 12: C++ allows us to throw whatever object as an exception, but it is nice to throw subclasses of `std::exception` (https://en.cppreference.com/w/cpp/error/exception). `std::runtime_error` is a very general one, and can fit all our cases. The issue on throwing objects that are not subclasses of `std::exception` is that if we error out on an exception (e.g. the given fasta/q file does not exist), C++ will error out with this message (in this example, as we are throwing `std::string`):
```
terminate called after throwing an instance of 'std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >'
```
`std::string` is an alias for `std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >`. Anyway, all that C++ says is that an exception of type `std::string` or whatever other type was thrown, but it does not describe the error itself. If we instead `throw std::runtime_error("Unable to open " + filepath);`, we get this when a file does not exist:

```
terminate called after throwing an instance of 'std::runtime_error'
  what():  Unable to open reads.fa
```

The first line is identical as before, just outputting the type of the exception, but then C++ sees that the object thrown is an `std::exception`, and calls the `what()` method, which prints the string given to the object at its construction. Then the user can understand the error.

You could also build your own exception classes for each type of exception your code might throw, but I don't think this is necessarily required here... just throwing an object that is a `std::exception`, either from std library or your own class is enough

src/fastaq_handler.cpp
```diff
@@ -1,176 +1,83 @@
...
     }
-}
-
-void FastaqHandler::skip_next()
-{
-    // cout << "skip ";
-    if (!line.empty() and (line[0] == '>' or line[0] == '@')) {
-        ++num_reads_parsed;
+    if (read_status == -2) {
+        throw "Truncated quality string detected";
```
leoisl, line 33: Please throw a `std::exception` object instead (see comment above)

src/fastaq_handler.cpp
```diff
@@ -1,176 +1,83 @@
...
+    } else if (read_status == -3) {
+        throw "Error reading " + this->filepath;
```
leoisl, line 35: Please throw a `std::exception` object instead (see comment above)

leoisl (changes requested):
Just minor suggestions...
Did your real data test (i.e. getting the output of the code in this branch vs `dev` branch) yielded the same results? If so, I think implementation is fine!

src/fastaq_handler.cpp
mbhall88: Yep, totally agree. This was me being lazy.

mbhall88:
> Did your real data test (i.e. getting the output of the code in this branch vs dev branch) yielded the same results? If so, I think implementation is fine!

In the process of checking this now. I'll post when I have.

mbhall88:
Ok. The output it exactly the same as the tip of `dev` :tada:

leoisl (approved):

PR #234, opened 2020-09-26
Move variant discovery to its own subcommand

### LINKED ISSUE #228: Merge candidate regions
We need to merge candidate regions that are within a certain distance from each. For a thorough background/explanation of this problem, see https://github.com/mbhall88/head_to_head_pipeline/issues/53#issuecomment-694036442.

I guess a good starting default for the value of the distance would be k + 1

### LINKED ISSUE #233: Allow passing genome size string
It is a little cumbersome to pass the genome size as an integer. We should allow genome size strings like `4m`

mbhall88:
I found myself needing to expose more and more *de novo* parameters at the CLI and realised it was all cluttering `map`. There could also a *slightly* misleading perception that using `--discover --genotype` in `map` would genotype with the discovered variants (which it doesn't). Anyway, I forsee - when we have self-container *de novo* variant integration - the need to have this subcommand output a new PRG with variants added, or even another subcommand `augment`.

## Added
- New subcommand `pandora discover`
```
$ pandora discover --help
Quasi-map reads to an indexed PRG, infer the sequence of present loci in the sample and discover novel variants.
Usage: ./pandora discover [OPTIONS] <TARGET> <QUERY>

Positionals:
  <TARGET> FILE [required]    An indexed PRG file (in fasta format)
  <QUERY> FILE [required]     Fast{a,q} file containing reads to quasi-map

Options:
  -h,--help                   Print this help message and exit
  --discover-k INT            K-mer size to use when discovering novel variants [default: 11]
  --max-ins INT               Max. insertion size for novel variants. Warning: setting too long may impair performance [default: 15]
  --covg-threshold INT        Positions with coverage less than this will be tagged for variant discovery [default: 3]
  -l INT                      Min. length of consecutive positions below coverage threshold to trigger variant discovery [default: 1]
  -L INT                      Max. length of consecutive positions below coverage threshold to trigger variant discovery [default: 50]
  -P,--pad INT                Padding either side of candidate variant intervals [default: 22]
  -d,--merge INT              Merge candidate variant intervals within distance [default: 22]
  --min-dbg-dp INT            Minimum node/kmer depth in the de Bruijn graph used for discovering variants [default: 2]
  -v                          Verbosity of logging. Repeat for increased verbosity

Indexing:
  -w INT                      Window size for (w,k)-minimizers (must be <=k) [default: 14]
  -k INT                      K-mer size for (w,k)-minimizers [default: 15]

Input/Output:
  -o,--outdir DIR             Directory to write output files to [default: pandora_discover]
  -t,--threads INT            Maximum number of threads to use [default: 1]
  --kg                        Save kmer graphs with forward and reverse coverage annotations for found loci
  -M,--mapped-reads           Save a fasta file for each loci containing read parts which overlapped it

Parameter Estimation:
  -e,--error-rate FLOAT       Estimated error rate for reads [default: 0.11]
  -g,--genome-size STR/INT    Estimated length of the genome - used for coverage estimation. Can pass string such as 4.4m, 100k etc. [default: 5000000]
  --bin                       Use binomial model for kmer coverages [default: negative binomial]

Mapping:
  -m,--max-diff INT           Maximum distance (bp) between consecutive hits within a cluster [default: 250]
  -c,--min-cluster-size INT   Minimum size of a cluster of hits between a read and a loci to consider the loci present [default: 10]

Preset:
  -I,--illumina               Reads are from Illumina. Alters error rate used and adjusts for shorter reads

Filtering:
  --clean                     Add a step to clean and detangle the pangraph
  --max-covg INT              Maximum coverage of reads to accept [default: 600]

Consensus/Variant Calling:
  --kmer-avg INT              Maximum number of kmers to average over when selecting the maximum likelihood path [default: 100]
```
- function to merge candidate regions within a certain distance [closes #228] and `-d,--merge` to specify the distance at CLI
- `Interval::is_close` method
- `--min-dbg-dp` to expose minimum adbundance in GATB graph
- `-l` and `-L` for min/max candidate region length
- `-P,--pad` to specify candidate region padding

## Changed
- allow `-g,--genome-size` to take a string such as `4.4m` [closes #233]
- some logging cleanup/level changes
 
## Removed
- `--coverages` option from `map` as it wasn't being used

### DIFF
--- a/include/denovo_discovery/candidate_region.h
+++ b/include/denovo_discovery/candidate_region.h
@@ -87,26 +87,37 @@ private:
     Fastaq generate_fasta_for_denovo_paths();
 };
 
-std::vector<Interval> identify_low_coverage_intervals(
-    const std::vector<uint32_t>& covg_at_each_position,
-    const uint32_t& min_required_covg = 3, const uint32_t& min_length = 1,
-    const uint32_t& max_length = 76);
-
-using CandidateRegions = std::unordered_map<CandidateRegionIdentifier, CandidateRegion>;
-
-CandidateRegions find_candidate_regions_for_pan_node(
-    const TmpPanNode& pangraph_node_components,
-    const uint_least16_t& candidate_region_interval_padding = 0);
-
+using CandidateRegions
+= std::unordered_map<CandidateRegionIdentifier, CandidateRegion>;
 using ReadId = uint32_t;
-using PileupConstructionMap
-    = std::map<ReadId, std::vector<std::pair<CandidateRegion*, const ReadCoordinate*>>>;
+using PileupConstructionMap = std::map<ReadId,
+    std::vector<std::pair<CandidateRegion*, const ReadCoordinate*>>>;
 
-PileupConstructionMap construct_pileup_construction_map(
-    CandidateRegions& candidate_regions);
+class Discover {
+private:
+    const uint32_t min_required_covg;
+    const uint32_t min_candidate_len;
+    const uint32_t max_candidate_len;
+    const uint16_t candidate_padding;
+    const uint32_t merge_dist;
+public:
 
-void load_all_candidate_regions_pileups_from_fastq(const fs::path& reads_filepath,
-    const CandidateRegions& candidate_regions,
-    const PileupConstructionMap& pileup_construction_map, uint32_t threads = 1);
+    Discover(uint32_t, uint32_t, uint32_t, uint16_t, uint32_t);
+
+    std::vector<Interval> identify_low_coverage_intervals(
+        const std::vector<uint32_t>& covg_at_each_position);
+
+
+    CandidateRegions find_candidate_regions_for_pan_node(
+        const TmpPanNode& pangraph_node_components);
+
+
+    PileupConstructionMap pileup_construction_map(
+        CandidateRegions& candidate_regions);
+
+    void load_candidate_region_pileups(const fs::path& reads_filepath,
+        const CandidateRegions& candidate_regions,
+        const PileupConstructionMap& pileup_construction_map, uint32_t threads = 1);
+};
 
 #endif // PANDORA_CANDIDATE_REGION_H
--- a/include/denovo_discovery/candidate_region.h
+++ b/include/denovo_discovery/candidate_region.h
@@ -87,11 +87,10 @@ private:
     Fastaq generate_fasta_for_denovo_paths();
 };
 
-using CandidateRegions
-= std::unordered_map<CandidateRegionIdentifier, CandidateRegion>;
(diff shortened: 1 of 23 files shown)


mbhall88:
I have (successfully) run the tip of this branch on multiple samples.

include/denovo_discovery/candidate_region.h
```diff
@@ -87,26 +87,37 @@ class CandidateRegion {
...
+    const uint32_t min_candidate_len;
+    const uint32_t max_candidate_len;
+    const uint16_t candidate_padding;
+    const uint32_t merge_dist;
+public:
 
-CandidateRegions find_candidate_regions_for_pan_node(
-    const TmpPanNode& pangraph_node_components,
-    const uint_least16_t& candidate_region_interval_padding = 0);
+    Discover(uint32_t, uint32_t, uint32_t, uint16_t, uint32_t);
```
leoisl, line 105: Very minor: I have a preference to have the parameter names in the header file. Although it is not required from the compilation POV, the idea is that the header file presents the interface of a module/class, and hides the implementation from the user (which usually has no need to see the details), but when the parameters names are hidden in the header, we have to forcibly go to the implementation/cpp file to actually see what each parameter is. This is minor because this is just a few clicks away, and CLion I think takes care of this with inspections

src/interval.cpp
```diff
@@ -64,4 +62,37 @@ bool Interval::operator<(const Interval& y) const
...
+
+void merge_intervals_within(std::vector<Interval>& intervals, const uint32_t dist)
+{
+    if (intervals.size() < 2) { // nothing to merge
+        return;
+    }
+    std::sort(intervals.begin(), intervals.end());
+
+    // assumption that iv1 < iv2
+    auto close { [dist](const Interval& iv1, const Interval& iv2) -> bool {
```
leoisl, line 75: This looks like an useful method to add to the `Interval` class itself, we might need to use it elsewhere

leoisl (approved):
Just two minor comments, very nice PR, I think this was an essential refactoring, thanks for taking the time to do this

include/denovo_discovery/candidate_region.h
mbhall88: Ah ok. I also have the same preference, but I had noticed a lot of methods in the code base *didn't* name them so I was just trying to follow convention. However, I agree it is better to name them. Will fix this now.

mbhall88:
As a way of validating this is working - in tandem with https://github.com/mbhall88/head_to_head_pipeline/issues/54 - here is a picture of the precision and recall for the version of pandora in this PR

**fixed_prgs is the data from this PR**

## Recall

image

## Precision

Note: fixed_prgs is unfiltered here

image

leoisl (approved):

PR #250, opened 2020-11-25
Merging README updates to master

leoisl:


mbhall88 (approved):
For future reference, I think README updates can just be PR'd directly on to `master` as they don't change any code.

PR #255, opened 2021-01-07
Remove boost as dependency

### LINKED ISSUE #254: Remove boost and zlib as dependencies
Add boost as external project, making installation from source (and thus usage outside containers) much easier, requiring users only to have basic C++ environment (`g++`, `cmake`, `ld`, etc) and `zlib` pre-installed

leoisl:
Adds `boost` as external project, making installation from source (and thus usage outside containers) much easier, requiring users only to have basic C++ environment (`g++`, `cmake`, `ld`, etc) and `zlib` pre-installed. More specifically, in a `ubuntu:bionic` container, we just now need to install `build-essential git cmake wget zlib1g-dev` before compiling `pandora`.

I tried to remove `zlib` also as dependency, but unfortunately `pandora` dependencies (GATB, cgranges, and others) require a system-wide `zlib` installation... There should have ways to bypass this, but as `zlib` is such a standard library, I think it is fine to require it. Tell me if you think it is indeed necessary to remove also `zlib`.

There are also some CMake refactorings.

Closes #254

PS: I don't know if Travis CI build is working


CMakeLists.txt
```diff
@@ -78,38 +125,16 @@ file(GLOB_RECURSE SRC_FILES
         )
 
 add_executable(${PROJECT_NAME} ${SRC_FILES})
-add_dependencies(${PROJECT_NAME} gatb gtest)
+
+add_dependencies(boost gatb gtest)
+add_dependencies(${PROJECT_NAME} gatb gtest boost)
```
mbhall88, line 130: Is this repeated on purpose? (with the addition of the project name)

ext/boost.cmake
```diff
@@ -1,17 +1,19 @@
 include(ExternalProject)
 
-execute_process(COMMAND mkdir -p ${CMAKE_CURRENT_BINARY_DIR}/ext/boost_1_65_1)
+execute_process(COMMAND mkdir -p ${CMAKE_CURRENT_BINARY_DIR}/ext/boost_1_62_0)
```
mbhall88, line 3: Might be better to make a variable for the version number as we use it a few times in this file?

mbhall88 (approved):
Looks great. I think I mentioned somewhere that I think @rmcolq needs to give access from her Travis profile for the build to run there (or something like that - sorry just back from holidays)

mbhall88:
Also, given Travis doesn't seem to be running, have you tried this new install locally and also the container build?

CMakeLists.txt
leoisl: yeah, we could remove `gatb` and `gtest` from the second one, but I left just for documentation purposes... if we don't specify this, we have compilation issues that I don't exactly remember which one

leoisl:
> Also, given Travis doesn't seem to be running, have you tried this new install locally and also the container build?

yeah, locally and building in the container works!

leoisl:
There is a better solution using the [hunter](https://github.com/cpp-pm/hunter/) CMake package manager, will be closing this and reopening a new PR soon

Issue #262, opened 2021-02-11
Discussion: assumptions relating to cluster size expectations and coverage estimation


mbhall88:
> Disclaimer: I would like this issue to function as a type of "async" meeting.

I recently (re)read Rachel's thesis chapter relating to quasi-mapping and also walked through some of the code relating to this and have found some assumptions I would like to understand better. In particular, there are two assumptions I would like to challenge. I am not saying they are wrong, I would just like to see where the challenge takes us.

### Assumption 1: Setting the minimum cluster size

In setting the minimum cluster size (sub-heading in [Section 4.3.2][thesis]), we define the threshold for the number of hits in a cluster relative to the length of reads and local graphs. As these two lengths can differ from each other, we take the minimum of the two. However, we don't use the length of the read, we actually use the expected density of minimizers in a read of that length, which is `2||read||/w`. (I believe this comes from Section 3.1 of [this paper][density] - correct me if that is wrong Rachel).  
We then account for sequencing errors by multiplying this value by the `exp(-ke)`. Ultimately, this equation is 

![Screenshot from 2021-02-11 20-58-35](https://user-images.githubusercontent.com/20403931/107628131-ef019d80-6cab-11eb-91f7-2a49bb84c902.png)

where *l* is the minimum k-mer path length of the local graph and *s* is the sequence (read).

In reality, we don't quite do this though. The portion of the code that calculates the number of expected hits (prior to accounting for errors) is https://github.com/rmcolq/pandora/blob/815da22867f2bdaa3e3bf40f382be830c1506b0a/src/utils.cpp#L489-L498

and then later https://github.com/rmcolq/pandora/blob/815da22867f2bdaa3e3bf40f382be830c1506b0a/src/utils.cpp#L223-L227

This is where the assumption comes in. We assume that Nanopore reads are **always** longer than local graphs (`expected_number_kmers_in_short_read_sketch` is always set to uint_32 max for nanopore data). I appreciate that for *E. coli* datasets this is probably correct most of the time. But, for Mtb this is certainly *not* the case. I have plenty of samples with a median read length in the range 0.5-2kbp and lots of local graphs longer than that (by a lot in some cases).

**Question 1:** is this a fair assumption, and if not, should we replace it by checking this minimum value for each read?  
[...]

PR #265, opened 2021-03-11
Don't assume ONT reads are longer than loci

mbhall88:
A fix for the following issue.

(Taken from https://github.com/rmcolq/pandora/issues/262#issue-806308312)

In setting the minimum cluster size (sub-heading in [Section 4.3.2][thesis]), we define the threshold for the number of hits in a cluster relative to the length of reads and local graphs. As these two lengths can differ from each other, we take the minimum of the two. However, we don't use the length of the read, we actually use the expected density of minimizers in a read of that length, which is `2||read||/w`. (This comes from Section 3.1 of [this paper][density]).  
We then account for sequencing errors by multiplying this value by the `exp(-ke)`. Ultimately, this equation is 

Screenshot from 2021-02-11 20-58-35

where *l* is the minimum k-mer path length of the local graph and *s* is the sequence (read).

In reality, we don't quite do this though. The portion of the code that calculates the number of expected hits (prior to accounting for errors) is https://github.com/rmcolq/pandora/blob/815da22867f2bdaa3e3bf40f382be830c1506b0a/src/utils.cpp#L489-L498

and then later https://github.com/rmcolq/pandora/blob/815da22867f2bdaa3e3bf40f382be830c1506b0a/src/utils.cpp#L223-L227

This is where the assumption comes in. We assume that Nanopore reads are **always** longer than local graphs (`expected_number_kmers_in_short_read_sketch` is always set to uint_32 max for nanopore data). I appreciate that for *E. coli* datasets this is probably correct most of the time. But, for Mtb this is certainly *not* the case. I have plenty of samples with a median read length in the range 0.5-2kbp and lots of local graphs longer than that (by a lot in some cases).

From Rachel's (brilliant) thesis (emphasis mine)

> When the minimum size of a cluster is set too low, we have more false positive
local graphs identified as present in the dataset, and also have to handle more noise
downstream when inferring a mosaic sequence and genotyping. **When it is set too
high, we have less sensitivity to discover loci that are present.**

So, I think for Nanopore datasets with low median read lengths or very long loci, we could be losing sensitivity to discover loci.

This is backed up by the resulting plots in https://github.com/rmcolq/pandora/issues/262#issuecomment-778109077

[thesis]: https://ora.ox.ac.uk/objects/uuid:a8027db9-8d8c-4015-aab6-1371217ccb5c/download_file?file_format=pdf&safe_filename=RachelColquhounThesis.pdf&type_of_work=Thesis
[density]: https://theory.stanford.edu/~aiken/publications/papers/sigmod03.pdf


leoisl (approved):
All looks fine to me!

PR #266, opened 2021-03-11
Version 0.8.0

mbhall88:
Merging all changes related to the release of version 0.8.0


leoisl:
Ah, I have to update the [pre-compiled binary](https://github.com/rmcolq/pandora/releases/tag/v0.8.0-alpha) to include your [latest changes](https://github.com/rmcolq/pandora/pull/265). By the way, https://github.com/rmcolq/pandora/pull/257 makes lots of changes, some that were not tested. However, it is easy to spot if a bug was introduced (`pandora` will issue an error message and exits whereas it should not), but this requires testing it a bit further by running it. That is why we have [this pre-release](https://github.com/rmcolq/pandora/releases/tag/v0.8.0-alpha), and was thinking of using it for a few weeks before changing it to a release. By merging from `dev` to `master`, I guess this means that we are saying that this pre-release is actually stable and should become a release, as this is what users will get if they compile from source? Or do we keep v0.8.0 as a pre-release?

mbhall88:
If there's a bug we can just release a patch though. We are still on v0.x which means we are still saying the project is not stable. Point 4 from [SemVer](https://semver.org/)

> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

leoisl:
Alright, will update the precompiled binary, and the release!

leoisl:
Should probably set up a github action to do this on all tagged versions...

PR #268, opened 2021-03-17
Local updates with unit tests

leoisl:
Note: there is still a TODO at the end of this PR, but just requesting for comments for now;

This PR implements the changes needed for `pandora` to work with the [make_prg prototype that allows for local updates](https://github.com/leoisl/make_prg/).

# Changelog

There are several changes to the denovo component:

1. Like `pandora compare`, `pandora discover` now receives a read index files describing samples and reads, and discover denovo sequences in these samples. For this to be efficient, we use several cores, but we use multiprocessing instead of multithreading to bypass GATB issues with multithreading (this might be solved soon, see https://github.com/GATB/gatb-core/issues/37#issuecomment-774133658). Each sample is processed by a single-threaded child process from a pool. So if you specify 50 samples and 25 threads, each child process will process 2 samples using 1 thread. Note that `pandora map` CLI was not changed;
2. We don't output full ML sequences with the denovo variant applied anymore. We instead output, for each locus, the full ML path, and a VCF-like section to represent variants (see an [example here](https://github.com/leoisl/make_prg/blob/master/sample_example/denovo_paths/denovo_paths.txt)). These variants are obtained by aligning the ML sequence to its version with variants applied, and inferring where we have (multiple) SNPs and indels. This aligment is done with [SeqAn](https://github.com/seqan/seqan);
3. We don't create several files when running discovery anymore (one file per candidate region). We instead write all denovo paths information to a buffer, which is written to disk for each sample, and then we concatenate the file produced by each sample. So the final output is a single file;

A changelog with fewer details can be found [here](https://github.com/rmcolq/pandora/blob/local_updates_with_unit_tests/CHANGELOG.md#unreleased).

# Testing

1. Every code done in this PR is tested, except for the code in the `*_main.cpp` source files;
2. The hardest method to implement is [CandidateRegion::get_variants()](https://github.com/rmcolq/pandora/pull/268/files#diff-26955524ef07a048903706fe025a2d2dae0f2b8b4da6e3363a70baf1e3e2f9d3R249-R300). To ensure it is running fine, [40 unit tests were done for this method](https://github.com/rmcolq/pandora/pull/268/files#diff-8b770668aa01ae69ff2318262db80e8b83b306fe385b9f25a775839254f2bd73R1836-R2183);

# Usability

There is no need anymore to use a `snakemake` pipeline to run the full `pandora` pipeline, with denovo discovery. This can be done in 6 command lines (see an [example here](https://github.com/rmcolq/pandora/tree/local_updates_with_unit_tests/example)). Although in this example there is no need to have `pandora` or `make_prg` installed (the running script will automatically download and run the precompiled binaries for these tools), we still need `MAFFT`. [This issue](https://github.com/leoisl/make_prg/issues/1) concerns about removing this dependency.

# Performance

Notes:
1. This performance results are on the 20-way data, **nanopore** only;
2. Three genes were very slow, and were thus removed: `GC00000027_2`, `GC00000895_r1_r1_1`, `GC00004221`;
3. `pandora` was compiled from source and in `Release` mode, thus with high optimisations;
4. `make_prg` commit is `8575f3bc7f41fcbe678dcbcb373e6a867bbbeada`; `pandora` commit is `385a398ecb190dc2d45adc13ea6cb9a63d79a92f`;
5. All runs used 32 threads;

```
make_prg from_msa
 * time: 21m
 * RAM: < 1 GB
pandora index
 * time: 13m
 * RAM: 13.5 GB
pandora discover
 * time: 4h 50m
 * RAM: 10.5 GB
make_prg update
 * time: 1h 11m
 * RAM: 5 GB
pandora index (on updated PRG)
 * time: 1h 7m
 * RAM: 21.6 GB
pandora compare
 * time: 8h 7m
 * RAM: 25.1 GB
```

**Full pipeline stats**
**Total time: 15h 49m**
**Max RAM: 25.1GB**
**Huge improvements to development cycle.**

# Known issues

* Not all variants can be updated (minor bug that happens in edge cases):
```
Number of variants successfully applied: 980088
Number of variants that failed to be applied: 92
```

# TODO

* Add the results of this code with het filter on the 20-way. Show all plots, not only ROC curve.


mbhall88:
Great work!

A few things from me:
- Saying that `pandora discover` is multi-threaded is a little misleading. It is still single-threaded on a per-sample level. i.e. it isn't any faster than before this PR.
- I don't like the use of a proprietary output from `discover`. I think we should output VCF.

It's so awesome that the steps required to run the full cycle are simplified. However, you still do kind of need a pipeline to run this as there are 6 different steps involved. Maybe we should think about having "one subcommand to rule them all". For instance, `pandora genotype` which would run all 6 steps (or less if certain flags are set).

leoisl:
> * Saying that `pandora discover` is multi-threaded is a little misleading. It is still single-threaded on a per-sample level. i.e. it isn't any faster than before this PR.

You are right, will fix this.

> * I don't like the use of a proprietary output from `discover`. I think we should output VCF.

Yeah, the ideal is not output any proprietary format at all. We have to output two main information: the ML path through the PRG and the variants WRT this path. For the first one, we could use jVCF or GFA (not sure if GFA works), or keep it as a proprietary format, and for the second one we could use VCF. Producing VCF output now, with implementation, testing, and adapting `make_prg` to process it might take some 4-5 days. It is very doable and not a complicated issue, but I am not sure if we should close this development cycle and this PR so that we have a set version for the paper re-evaluation and review, and can start regenerate the results with this new version. We could switch to VCF format the first thing after this version. I think @iqbal-lab might know what to do here.

By the way, I changed my opinion that this PR has to be merged to `dev` or `master` since it will be in the paper. We can keep it in a branch, it is fine, as long as we can make a prerelease with it, so we can refer to this prerelease in the paper. 

> It's so awesome that the steps required to run the full cycle are simplified. However, you still do kind of need a pipeline to run this as there are 6 different steps involved. Maybe we should think about having "one subcommand to rule them all". For instance, `pandora genotype` which would run all 6 steps (or less if certain flags are set).

Yeah, we could add such commands! Within this context, I should also remember that we want `make_prg` to become a library, not an external executable. Then I think `pandora index` can receive a set of MSAs and 1) build and 2) index the PRGs, and `pandora discover` could discover the variants and update the PRGs and the index accordingly. Then we are back to 2 commands to run `pandora` without denovo discovery (`index` and `compare`) and 3 to run `pandora` with denovo discovery (`index`, `discover`, `compare`). Any or both of these options are fine to me.

iqbal-lab:
There's one point that has been slightly lost in the discussion above, about whether we should be outputting stuff in VCF.

The output from de novo discovery for a single sample and gene looks like this

2 samples
Sample toy_sample_1
1 loci with denovo variants
GC00010897
9 nodes 
(0 [0, 110) ATGCAGATACGTGAACAGGGCCGCAAAATTCAGTGCATCCGCACCGTGTACGACAAGGCCATTGGCCGGGGTCGGCAGACGGTCATTGCCACACTGGCCCGCTATACGAC)
(1 [113, 114) G)
(3 [121, 171) GAAATGCCCACGACCGGGCTGGATGAGCTGACAGAGGCCGAACGCGAGAC)
(4 [174, 175) G)
(6 [182, 301) CTGGCCGAATGGCTGGCCAAGCGCCGGGAAGCCTCGCAGAAGTCGCAGGAGGCCTACACGGCCATGTCTGCGGATCGGTGGCTGGTCACGCTGGCCAAGGCCATCAGGGAAGGGCAGGA)
(7 [304, 308) ACTG)
(9 [319, 360) CGCCCCGAACAGGCGGCCGCGATCTGGCACGGCATGGGGGA)
(10 [364, 365) G)
(12 [374, 491) GTCGGCAAGGCCTTGCGCAAGGCTGGTCACGCGAAGCCCAAGGCGGTCAGAAAGGGCAAGCCGGTCGATCCGGCTGATCCCAAGGATCAAGGGGAGGGGGCACCAAAGGGGAAATGA)
2 denovo variants for this locus
44	C	T
422	A	T


The stuff at the end could be put in a VCF with respect to the ML path (which is specific to that sample and gene).
**However**, the code needs those last 2 lines AND the 17 lines before them too, and it interprets the coordinates in terms of both the bracketed stuff and the sequence. Our graph is proprietary and we necessarily have 17/19 lines which are pandora-specific, for this internal step (the file is not intended for the user). So, if we do split our 1 file into two, a VCF and one with proprietary stuff, we will then need to open both files, and parse them both in parallel. I don't think it makes sense to do it on that basis. Although we can use VCF to hold _some_ of the info, our process is very tied to the different paths through the graph, not just vCF-encodable stuff.

mbhall88:
> By the way, I changed my opinion that this PR has to be merged to dev or master since it will be in the paper. We can keep it in a branch, it is fine, as long as we can make a prerelease with it, so we can refer to this prerelease in the paper.

Why not just move the paper tag to this commit?

Sorry for all the confusion re: outputting VCF. I've just realised the key bit of info that I was missing what the position of the ML path within the PRG - i.e. the thing that allows the local updates in the first place. This could be encoded in the INFO field of a VCF though no? There just seems to be a bit of redundancy in that intermediate file. We have the sequence of the ML path already in the consensus fastq. So the ML path coordinates that correspond to the location within the PRG could just be encoded in each denovo variant't INFO field no? Do we really need *all* of the ML path -> PRG coordinates?

leoisl:
> Why not just move the paper tag to this commit?

Yeah, will do that, but I also want to have a prerelease to provide the precompiled binary of `pandora` for the version we used in the paper. Need a prerelease to upload this asset.

> This could be encoded in the INFO field of a VCF though no?

yes

> So the ML path coordinates that correspond to the location within the PRG could just be encoded in each denovo variant't INFO field no?

yes

> Do we really need _all_ of the ML path -> PRG coordinates?

No. I've put all coordinates because we are describing all variants of the ML path in this file format (e.g. if we have 100 denovo variants, we just need to specify all coordinates once and we are done). But if we will instead put the coordinates in the INFO field, then if we have 100 denovo variants, we would need to specify all coordinates 100 times, which is big waste of space. In this case, I think that for each variant, we can just specify the coordinates surrounding the variant, and it should work fine.

PS: after writing this, it is definitely possible to specify just the coordinates surrounding the variant in the INFO field, but the implementation will become more complicated. However, we have the ML path sequence already in the consensus fastq. I guess we could just describe the ML path itself in the header of the sequence? Then the consensus fastq gives us the ML paths and sequences, and the VCF gives us the variants on these paths, and we have exactly the same info as the current proprietary format with two very standard formats, and with minimal adaption effort.

iqbal-lab:
I'm not very keen on putting complex stuff in the sequence header, which could get very big for long genes.
I also think the fasta is only an implicit path, it's much less explicit than the current proprietary format. I don't think it makes sense to push for standard formats for internal temporary files unless they are appropriate for what we are storing. The information in the current file is really quite specific and not what VCF is meant to hold.  At the moment i don't see an argument for shifting to vCF

leoisl:
Ok!

iqbal-lab:
Maybe we need to step back and remind/ask ourselves what this is for - am basically checking myself, to see if I am arguing unnecessarily/wrongly. Let's see - is this right?

 We want to pass information on SNPs/indels on where on the ML path they are, and what the ML path is, such that it can be used to place this list of variants in "the rectangle of the MSA", or "the leaves of the make-prg recursion tree". 
@mbhall88 is completely right that VCF is perfect for holding variants with respect to the ML path. 
As i understand it @leoisl , currently you use the PRG encoding of the ML path, plus the coord along the path, to transfer information on where in the rectangle the variant is. Is this right?

leoisl:
> As i understand it @leoisl , currently you use the PRG encoding of the ML path, plus the coord along the path, to transfer information on where in the rectangle the variant is. Is this right?

Exactly

mbhall88:
> > Why not just move the paper tag to this commit?
> 
> Yeah, will do that, but I also want to have a prerelease to provide the precompiled binary of `pandora` for the version we used in the paper. Need a prerelease to upload this asset.

You can upload anything to a tag. For example, for the previous paper tag <https://github.com/rmcolq/pandora/releases/tag/pandora_paper_tag1> you click "Edit tag" and then you can attach any file you want to it.

mbhall88 (approved):
Ok, well I've made my thoughts clear regarding the work in this PR. I know VCF as an intermediate is more work now but I think it will create a lot less work in the future. Proprietary formats always have higher upkeep costs.  
I will leave it there and stop holding this PR up.

leoisl:
> You can upload anything to a tag. For example, for the previous paper tag https://github.com/rmcolq/pandora/releases/tag/pandora_paper_tag1 you click "Edit tag" and then you can attach any file you want to it.

Ha! I did not know that! Ok, that is fine with me then. We just move the paper tag to this commit and we upload the precompiled binary to the paper tag.



> Ok, well I've made my thoughts clear regarding the work in this PR. I know VCF as an intermediate is more work now but I think it will create a lot less work in the future. Proprietary formats always have higher upkeep costs.
> I will leave it there and stop holding this PR up.

There might be no need to merge this PR yet. We can just move the paper tag to this commit and upload the precompiled binary, and it is all fine by me. Then once we are totally happy with this PR, regarding both `pandora` and `make_prg` implementations, and file formats, we can merge. I actually now prefer to not do the merge, as `make_prg` is still a prototype and lacks unit tests. I guess a merge to `dev` is fine, but definitely not to `master`. Anyway, we can keep this PR open until we have everything correct also. I don't have any hard constraints at all if we should merge or not, as you solved my constraints. I will leave the decision to you and Zam

iqbal-lab:
I think the argument that proprietary formats make more work is a perfectly good rule of thumb, but simply doesn't work here. No standard file format does what we need, I think we'd be worse off with vcf. If you have a need, and the standard doesn't do what you want, you shouldn't be completely trapped by the standard. Especially for this internal temporary file not for sharing. The key data we want to convey is which  subrectangle of the MSA the variant is in. You can't do that with vcf

leoisl:
> You can upload anything to a tag. For example, for the previous paper tag https://github.com/rmcolq/pandora/releases/tag/pandora_paper_tag1 you click "Edit tag" and then you can attach any file you want to it.

This sort of worked, but not the way we wanted. When we edit a tag, we are taken into the release page. So I released the version in this branch as "Version used in pandora paper" and tagged as pre-release (see https://github.com/rmcolq/pandora/releases). Not sure if we should keep it like this though...

iqbal-lab:
AFAIK this is then ready to merge to dev?

leoisl:
If it is, we can follow these steps next:

1. Merge this branch to `dev`;
2. Make prerelease and provide binary for the paper;
3. Add unit tests to `make_prg` prototype and turn it into a production version;
4. Release both `pandora` and new `make_prg`;

Note that we can't release the version of `pandora` in this branch without releasing `make_prg`, as this version is only compatible with `make_prg` prototype, that has the `update` feature and other changes. This version of `pandora` is not compatible with the current releases of `make_prg` in https://github.com/iqbal-lab-org/make_prg . But I think if we tag the release as pre-release, there will be no issue.

I wonder if @mbhall88 agrees with this or suggests another path

mbhall88:
> I wonder if @mbhall88 agrees with this or suggests another path

I already approved the PR?

PR #271, opened 2021-04-01
0.9.0 -> 0.9.0-rc1

leoisl:
Change 0.9.0 -> 0.9.0-rc1.

Also has a small fix to allow us to put labels on versions (`CMake` does not allow labels on versions, just major, minor and patch versioning, and these have to be numbers).


.gitignore
```diff
@@ -130,3 +130,5 @@ example/venv
 /example/pandora-linux-precompiled-v0.9.0.gz
 
 example/pandora-linux-precompiled-pandora_paper_tag1
+
+example/pandora-linux-precompiled-v0.9.0-rc1
```
mbhall88, line 134:
```suggestion
example/pandora-linux-precompiled*
```

CHANGELOG.md
```diff
@@ -9,7 +9,7 @@ project adheres to
 
 ## [Unreleased]
 
-## [0.9.0]
+## [0.9.0-rc1]
```
mbhall88, line 12: When you wrap this in square brackets and don't add a link after you need to make sure there is a defined link defined elsewhere. i.e. check near the bottom for examples, create a link `[0.9.0-rc1]: <URL>`

mbhall88 (changes requested):

.gitignore
leoisl: I actually ended up cleaning this `.gitignore` file as there were so many outdated entries. Could you please re-review?

CHANGELOG.md
leoisl: Yep, sorry for missing that! Fixed this in the `CHANGELOG.md` and in the pre-release description. Also fixed a small issue with the `Unreleased` link, could you please re-review?

mbhall88 (approved):

PR #272, opened 2021-04-07
0.9.0-rc2

leoisl:
### Changed
- `pandora discover` now processes one sample at a time, but runs with several threads on the heavy tasks, i.e. when
mapping reads, finding candidate regions, and finding denovo variants. The result is that it now takes a lot less RAM to
run on multiple samples.

[0.9.0-rc2]: https://github.com/rmcolq/pandora/releases/tag/0.9.0-rc2

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,13 @@ project adheres to
 
 ## [Unreleased]
 
+## [0.9.0-rc2]
+
+### Changed
+- `pandora discover` now processes one sample at a time, but runs with several threads on the heavy tasks, i.e. when
+mapping reads, finding candidate regions, and finding denovo variants. The result is that it now takes a lot less RAM to
+run on multiple samples at the cost of a small increase of runtime.
+
 ## [0.9.0-rc1]
 
 ### Changed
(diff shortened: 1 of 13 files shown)


mbhall88 (approved):
Given we are tagging this stuff (even though as a pre-release) I think it should be on `master`. Otherwise, it becomes very confusing the code people see when coming to the repo (i.e. `master`) is different to the latest tagged code (i.e. `dev`).

leoisl:
I am fine with merging this to `master`, as it has unit tests, but the issue is that this version of `pandora` depends on a `make_prg` prototype that is untested. This is actually the only reason I am marking these releases as pre-releases, and I am a bit hesitant to merge to `master`. As soon as the paper stuff is finished, I will add unit tests and thoroughly test the `make_prg` prototype, transforming it into a production version, and then I think we can do actual releases and merge to `master`. If the paper results on this new version of `pandora` and the `make_prg` prototype turns out fine, it might be ok to merge to `master` already, but I'd leave this decision to you and Zam then. My POV is that `dev` can be ahead of master and can have unstable versions, but master should always point to the latest stable one. OFC this is totally arguable, what do you think?

iqbal-lab:
I think we all agree with this 
>My POV is that dev can be ahead of master and can have unstable versions, but master should always point to the latest stable one. 

The question is , what counts as tested and stable. If the new make-prg is giving "the same" results as the old one with the 20way, then that is, at a system level, quite a lot of testing of 37,000 different instances of make-prg.
Personally i'm ok with merging to master, and then adding unit tests to make-prg, and if issues come up, fixing them.

iqbal-lab:
Has Brice tried out the new make-prg?

leoisl:
> Has Brice tried out the new make-prg?

No, as far as I am aware

leoisl:
As of now, I actually just need to monitor the 20-way pipeline and ensure it runs fine. OFC this is a very idle task, so I am either resuming my try to make a `pandora` conda recipe or I can start already adding unit tests to `make_prg`. Which of these tasks you think I should prioritize?

mbhall88:
Given all of this, I vote to merge to `master`. As these tagged releases are clearly stated as pre-releases it's fine if there is minor instability - so I'm happy with it being on `master`.

iqbal-lab:
I'm also happy

leoisl:
Alright, will be merged to `master`!

PR #274, opened 2021-04-18
Make pandora instalable through conda

leoisl:
* This PR includes a script, `scripts/create_archives.sh`, that will create `.tar.gz` and `.zip` packages of the `pandora` source code to be added to the releases. Differently from the source code packages automatically created by github, these packages actually include _all_ source code, including those from git submodules (it is weird why github does not include code from submodules directly. Several people call this actually a github bug, and an issue has been filled to github devs to fix this... but IIRC, the issue is still being fixed for a couple of years...);
* It also includes some few, but necessary, changes to the `CMake` files so that it is compilable in the conda build environment. The changes are very small, but it took a long time to find them... C++ tooling inside the conda container build env is something I really didn't appreciate... C++ tooling is already non trivial, and conda does not have the expected gcc/g++ toolsets in the default locations, which complicate things further. Anyway, the solution was actually **way** simpler than I expected, but I was unlucky to find it just after a couple of days playing with the conda build environment. Anyway, it is done now.
* Will convert this draft PR to a real PR once I finish doing the last checks, and check if the conda recipe is indeed correct (I can build it fine locally, unsure if can build it in their circle CI);

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -118,3 +118,5 @@ build_portable_executable
 pandora-linux-precompiled
 
 /cmake-build-release/
+/example/
+/archives/
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -98,6 +98,7 @@ set(Gtest_LIBRARIES GTest::gtest GTest::gmock_main)
 ########################################################################################################################
 # INSTALL BOOST
 if(BIOCONDA)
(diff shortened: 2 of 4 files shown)


leoisl:
Closing, and reopening where the fork origin is `rmcolq` due to technical reasons

PR #284, opened 2021-10-06
Change SVTYPE to VC

### LINKED ISSUE #249: Incorrect usage of SVTYPE
The INFO tag `SVTYPE` is reserved for describing structural variants (see [specs](https://samtools.github.io/hts-specs/VCFv4.3.pdf) section 3). However, we use it to describe variant types instead. This could lead to problems with downstream VCF processing tools.

As an example, `cyvcf2` uses the presence of that tag to indicate that a record is a SV. So, for a pandora VCF, every record would be considered an SV.

mbhall88:
### Changed
- The VCF INFO field `SVTYPE` has now been changed to `VC` [closes #249] *a la* [dbSNP](https://www.ncbi.nlm.nih.gov/snp/docs/products/vcf/redesign/#var_type)

Note: `dev` is obviously behind `master` so you can ignore the commits not from me.


mbhall88:
closing in favour of #285

PR #285, opened 2021-10-20
Robust TSV file parsing

### LINKED ISSUE #213: tsv file for pandora compare
This is not a big issue, but maybe you could mention somewhere that the tsv file you give to `pandora compare` needs an empty line at the end? I was very confused because I only got 26 out of my 27 samples compared, until I added a new line at the end of my list.

### LINKED ISSUE #249: Incorrect usage of SVTYPE

mbhall88:
### Fixed

- More robust TSV file parsing. Empty line no longer required at end and malformatted lines raise error [closes #213]

This PR also supersedes #284 which does the following

### Changed
- The VCF INFO field `SVTYPE` has now been changed to `VC` [closes #249] *a la* [dbSNP](https://www.ncbi.nlm.nih.gov/snp/docs/products/vcf/redesign/#var_type)


leoisl (approved):
All looks good!

PR #295, opened 2022-09-12
Handle ambiguous bases

### LINKED ISSUE #294: How to deal with ambiguous bases in reads
I have a fastq that gets zero coverage on all PRGs - which is very weird as there are lots of reads.

Looking at the log, pretty much every read has this warning

```
bad letter - found a non AGCT base in read so skipping read SRR1723506.87
```

Looking at the fastq, indeed, nearly every read has at least one `N` in it.

Is there a way we can deal with these?

For context, I came across this while testing `drprg`. mykrobe and tbprofiler make the correct predictions for this sample, but pandora doesn't even produce a genotyped VCF as it filters out all PRGs as they all fail filtering. 

How does cortex deal with this sort of thing @iqbal-lab ?

mbhall88:
This PR closes #294. I also brings `master` up to date with `dev`.

The methodology here is that we split a `Seq`'s internal sequence into a vector of strings that don't contain ambiguous bases - i.e., splitting the original sequence on ambiguous bases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Fixed
 
 - More robust TSV file parsing. Empty line no longer required at end [[#213][213]]
+- Handle ambiguous bases properly instead of skipping to next readonce we reach one [[#294][294]]
 
 ## [0.9.1]
 
@@ -134,5 +135,6 @@ their changes meticulously documented here.
 [234]: https://github.com/rmcolq/pandora/pull/234
 [249]: https://github.com/rmcolq/pandora/issues/249
 [265]: https://github.com/rmcolq/pandora/pull/265
+[294]: https://github.com/rmcolq/pandora/issues/294
 [v0.7.0]: https://github.com/rmcolq/pandora/releases/tag/v0.7.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Fixed
 
 - More robust TSV file parsing. Empty line no longer required at end [[#213][213]]
-- Handle ambiguous bases properly instead of skipping to next readonce we reach one [[#294][294]]
+- Handle ambiguous bases properly instead of skipping to next read once we reach one [[#294][294]]
 
 ## [0.9.1]
 
--- a/include/seq.h
+++ b/include/seq.h
@@ -11,9 +11,12 @@ class Seq {
 public:
     uint32_t id;
     std::string name;
-    std::vector<std::string> seq;
-    std::vector<size_t> offsets;
     std::set<Minimizer> sketch;
+
+    // the original sequence is split into several valid subsequences (composed of ACGT only)
+    std::vector<std::string> subseqs;  // these are the subsequences themselves
+    std::vector<size_t> offsets;  // these are the subsequences offsets on the original string
+
 
     Seq(uint32_t, const std::string&, const std::string&, uint32_t, uint32_t);
 
--- a/include/seq.h
+++ b/include/seq.h
@@ -25,7 +25,7 @@ public:
     void initialize(
         uint32_t, const std::string&, const std::string&, uint32_t, uint32_t);
 
-    bool add_letter_to_get_next_kmer(const char&, const uint64_t&, const uint64_t&,
+    void add_letter_to_get_next_kmer(const char&, const uint64_t&, const uint64_t&,
         uint32_t&, uint64_t (&)[2], uint64_t (&)[2]);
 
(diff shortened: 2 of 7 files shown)


mbhall88:
Okay, so I've implemented the first part of what you suggested @leoisl - i.e., split the sequence string into a vector of strings with no ambiguous bases.

**However**, the issue we face now (the failing test in `seq_test.cpp`) is that we lose the position of each minimizer on the read...

leoisl:
[Last commit](https://github.com/rmcolq/pandora/pull/295/commits/db2b135a6d6e6f62955fbf3b8bc6a64c6fae9996) records the offsets of each non-ambiguous substrings so that we can recover the position of the minimizers on the original read. All tests passing. Could you please check? Would like to do a round of refactoring before merging though...

mbhall88:
Alright, I've added some more tests and tested it on the sample in #294. 

The example read file from #294 still fails to map anything though. But I think this is now due to the decreased cluster size when filtering hits. For example, this read
```
AACTCGCGCGCGCCGAGCTGACCGACATNNNNCGGCTCACCGGCGAGATCAACACCCTGGCCCAGC
```
There are 5 minimizers (8 total, but there are some duplicates and the sketch is stored as a set) as a result of the Ns in the middle wiping out a bunch of space that would be filled by minimizers. Our default minimum cluster size (`-c`) is 10, so we filter out these minmizers. The same thing happens for all of the other reads.

This is somewhat of a corner case as this sample was sequenced on an Illumina Genome Analyzer II, which is quite old. But it is still informative I guess.

> Would like to do a round of refactoring before merging though...

Sure. What were you thinking of refactoring? Feel free to refactor whatever you like.

mbhall88:
Alright, I've added some more tests and tested it on the sample in #294. 

The example read file from #294 still fails to map anything though. But I think this is now due to the decreased cluster size when filtering hits. For example, this read
```
AACTCGCGCGCGCCGAGCTGACCGACATNNNNCGGCTCACCGGCGAGATCAACACCCTGGCCCAGC
```
There are 5 minimizers (8 total, but there are some duplicates and the sketch is stored as a set) as a result of the Ns in the middle wiping out a bunch of space that would be filled by minimizers. Our default minimum cluster size (`-c`) is 10, so we filter out these minmizers. The same thing happens for all of the other reads.

This is somewhat of a corner case as this sample was sequenced on an Illumina Genome Analyzer II, which is quite old. But it is still informative I guess.

> Would like to do a round of refactoring before merging though...

Sure. What were you thinking of refactoring? Feel free to refactor whatever you like.

mbhall88:
I've also made a binary https://github.com/rmcolq/pandora/pull/295/commits/f3662695568edc9d251363dae25e9cbfecb1f9a0 (current tip of this PR) of here: https://github.com/mbhall88/pandora/releases/tag/0.9.2-rc.1

iqbal-lab:
I think it's 100% acceptable to entirely lose short reads

mbhall88:
> I think it's 100% acceptable to entirely lose short reads

I would normally agree, but the machine this was sequenced on states

> Yield greater than 85% of bases higher than Q30 at 2 x 50 bp

So, all the reads are 50bp. So we'd potentially throw away the entire file when the bases that *are* there are **very** good.

I'm definitely not saying we should change the default minimum cluster size though. But it's good this is documented in case others come across it.

iqbal-lab:
Personally i'm ok to lose high quality 50bp read files.

iqbal-lab:
It's hard enough supporting nanopore and good/modern illumina, I wouldn't want to expend dev effort or make code more complex to handle legacy data.  I think it would even be ok to explicitly say we target reads longer than 99bp

mbhall88 (commented):
I think we pull that commit. It is way longer, I would argue harder to read, and brings boost into the picture without any real benefit…

Feel free to benchmark the two implementations though? Any differences will be quite important as we are running this over every base in the fastq.

leoisl:
> I think we pull [07048fb](https://github.com/rmcolq/pandora/commit/07048fbfaf7c79e4752dd09502729b750cf46972). It is way longer, I would argue harder to read, and brings boost into the picture without any real benefit…

Agreed

> Feel free to benchmark the two implementations though? Any differences will be quite important as we are running this over every base in the fastq.

I am sure your implementation is as fast as we can get

leoisl:
I've just added 3 commits. 2 refactoring, mostly variable renaming: https://github.com/rmcolq/pandora/pull/295/commits/2e5f583fa1bd0b356e907c16cce1ebbb82eb735e and https://github.com/rmcolq/pandora/pull/295/commits/a5eb55dec4ad911a992f4bff36a30686df7c8d6e. And 1 that basically errors out if we find an ambiguous base in `Seq::add_letter_to_get_next_kmer()` (we should not have ambiguous bases anymore here): https://github.com/rmcolq/pandora/pull/295/commits/f14cbcec0f8961b00341225ba50793bb3c4a5b9d

leoisl (approved):

leoisl:
PR fine to be merged by me, please check if you agree with the changes. If yes, feel free to merge. Otherwise, we can discuss further.

PR #299, opened 2022-10-17
Use racon for de novo variant discovery

mbhall88:
**TODO**
- [x] Sort out the merge conflicts
- [x] Define SAM file format (https://github.com/rmcolq/pandora/issues/300)
- [x] Finish producing the debugging extra files (we create four extra files when mapping for each sample: a minimizer PAF file, a .minimatches file, and two files describing the clusters definition and filtering. These are debugging files, they seem very useful, but there is no need to thorough unit test. Their production will be hidden by default, and can be activate just with an env variable);
- [ ] Check new index param: indexing_upper_bound;
- [ ] Check and test `racon` discovery;
- [x] Test out the rectified version for clear runtime bugs
- [ ] Write any further unit tests
- [ ] Update documentation


mbhall88:
The tip if this branch doesn't currently build.

The first error I came across is located at https://github.com/rmcolq/pandora/blob/629a134ac7efb15c6dba2e1e3d02da670efac486/src/minimizermatch_file.cpp#L19

We changed the `Seq` class such that string `Seq.seq` is now a vector `Seq.subseqs`. So I guess the correct way to fix this is to add a method `Seq.substr`. I would have implemented this, but I am just a little unsure of the behaviour if we request a substring that spans a "gap" segment - i.e. a segment where the read had an ambiguous base?

The other question I have is related to the same line. The variable is called kmer, but it seems the substring we are extracting is not necessarily a kmer in the sense of using the kmer value pandora uses... it just seems to extract the sequence from the read start index to the end of the PRG? I am a little confused. Should it not extract up until the read end position variable?

leoisl:
> We changed the `Seq` class such that string `Seq.seq` is now a vector `Seq.subseqs`. So I guess the correct way to fix this is to add a method `Seq.substr`. I have started implementing this method in [344a4fe](https://github.com/rmcolq/pandora/commit/344a4fe4b057903dfd496299abb036025ca64750), but I am just a little unsure of the behaviour if we request a substring that spans a "gap" segment - i.e. a segment where the read had an ambiguous base?

I went for another approach. Indeed, for `pandora` in the main branch, substrings from the read composed of ambiguous bases adds no value. However, I think this is not the case for an external read polisher and the SAM file we are producing in this branch. I actually don't know if `racon` makes use of the information in ambiguous bases, but surely makes use the full read length information (i.e. I don't think breaking the original read into subreads of valid bases would be good for `racon` as it is ok for `pandora`). Also the SAM file we produce represents information on the span of the whole read. Given these two motivations, I added the `Seq::full_seq` attribute back to the `Seq` class (this was the old `Seq::seq` attribute, representing the original sequence, with ambiguous bases). I think that now that we have two modules that need to refer to the original sequence, it makes sense to plug it back.

Let's call the previous approach `approach 1`. An alternative approach (`approach 2`) would be to reconstruct the original sequence from `Seq::subseqs` and `Seq::offsets`. This is partially possible, as we don't know specifically which ambiguous bases we have, but we could always join the subseqs together with `N`s. But that would be quite a significant amount of work instead of just storing the original sequence in `approach 1`. `approach 2` also uses half the RAM of `approach 1`, as it does not store the original sequence. However, we just keep a small amount of reads in RAM when mapping (one batch of 1000 reads per mapping thread), so I don't think this increase in RAM is significant. Besides, we can optimise the RAM usage in `approach 1` by storing the `Seq::subseqs` as `std::vector<std::string_view>` of the original sequence, and thus not wasting any RAM (this is just noted as a TODO now as I don't think this improvement is significant).

Given that we restored the ability to store the full read back to the `Seq` class, the code now trivially compiles and the tests passes. As I am not sure you will accept this other approach, I did this in [this commit](https://github.com/rmcolq/pandora/commit/c48dd8c87e3965d45f01c28316cfed5fd00d9d18),  in [this branch](https://github.com/rmcolq/pandora/tree/racon_denovo_p1).




> The other question I have is related to the same line. The variable is called kmer, but it seems the substring we are extracting is not necessarily a kmer in the sense of using the kmer value pandora uses... it just seems to extract the sequence from the read start index to the end of the PRG? I am a little confused. Should it not extract up until the read end position variable?

This writes to a `*.minimatches` file that contains all minimiser matches and other info. It extracts the subsequence starting from the starting position of the hit in the read for the length of the hit path in the PRG. Hit lengths in `pandora` is always the size of a `kmer`, so `hit->get_prg_path().length()` is equivalent to the kmer size. This is a few lines of such file:

```
kmer    read    read_start      read_end        read_strand     prg     prg_strand      prg_path
TGGCACGGCATGGGG simulated_read_0        5       20      1       GC00010897      1       1{[343, 358)}
ACGGCATGGGGGAGG simulated_read_0        9       24      1       GC00010897      1       3{[347, 360)[369, 370)[374, 375)}
TGCCGACCTCCCCCA simulated_read_0        15      30      0       GC00010897      0       3{[353, 360)[369, 370)[374, 381)}
```

mbhall88:
No, approach 1 sounds totally acceptable. As you say, we can optimise memory if it becomes an issue, but I doubt it will be a problem.

> Hit lengths in pandora is always the size of a kmer, so `hit->get_prg_path().length()` is equivalent to the kmer size.

Ah ok, cool.

leoisl:
Several fixes in the last commits:
* Racon denovo code was reviewed with some fixes;
* `pandora map` command is working again;
* `--keep-extra-debugging-files` option added to pandora `map`, `discover` and `compare`;

Tomorrow I can finish all the remaining points, then will move to testing.

leoisl:
Changes today:
* Some refactoring;
* Added a PAF-like file for minimizers (describing minimizer alignment between reads and PRGs);
* Fixed some issues with SAM file, is now compliant with the spec, and we can run for e.g. `samtools` in a `pandora` SAM file;

TODO:
* Fix strand flag in SAM file;
* Fix tests;

mbhall88:
Uploading a precompiled binary of https://github.com/rmcolq/pandora/pull/299/commits/1f9f7fbfea0c24997cba165cb123050e2a51e082 for debugging and testing (so there is a URL I can download from in drprg)
[pandora-1f9f7fb.gz](https://github.com/rmcolq/pandora/files/9834729/pandora-1f9f7fb.gz)

leoisl:
Tests are now compiling and passing, but `racon` discovery has 0% coverage. Can get it to 100% coverage on monday I think

mbhall88:
Alright, see https://github.com/mbhall88/drprg/issues/13#issuecomment-1294294433. I have tested this branch out on a couple of previously failing isolates - by failing I mean pandora didn't find novel variants - and it works perfectly for them.

I will test on a few more, but I think it's safe to say this PR is good to go if you want to merge @leoisl? Unless you're still working on something? We can do an alpha release and work a bit more on the SAM stuff and anything else we need to do to make this stable.

mbhall88:
Yeah, I guess that's the point of alpha releases though, "we are actively developing the features that will be released in 0.10.0” - no user would use it expecting it to be stable.

But it probably does make sense to add unit tests to this PR as they're one and the same with the added feature of denovo with racon

mbhall88:
So I've run on more samples now and it is working really well!

My only comment is regarding runtime. I am seeing increases in runtime of a few minutes per sample. Not a huge problem for drprg really. But I noticed we run racon for every locus? Is that necessary? Could we somehow leverage the previous candidate discovery stuff to only run racon on those loci where we think there might be novel variants?

This is more of a big picture question, definitely not suggested this is something that needs to be done for this PR.

leoisl:
> My only comment is regarding runtime. I am seeing increases in runtime of a few minutes per sample.

Would you know the runtime of `pandora` with DBG denovo and with `racon`? IDK if a few minutes more is a big or small proportion of the total runtime. Also, did you run with the `--debugging-files` flag on or off? If it is on, I'd say this would contribute to a big increase of runtime, as we are outputting lots of data to understand exactly what `pandora` does in each mapping and discovery step, and writing to disk is very slow (this increase in runtime is fine, as `--debugging-files` should be on only if you want to debug what `pandora` is doing).

> we run racon for every locus? Is that necessary? Could we somehow leverage the previous candidate discovery stuff to only run racon on those loci where we think there might be novel variants?

We could indeed have improvements in this area. We actually run `racon` for every locus several times, for a max of 10 rounds. We have an early stop mechanism if we see that `racon` did not improve the sequence, but in the worst case we might need to run 10 rounds. The improvements we did in this area is that neither `minimap2` nor `racon` are called as external process, but is just a C++ function call, i.e. we use them as APIs, so we avoid all the overhead of creating processes, terminating and cleaning them up etc (e.g. see https://github.com/rmcolq/pandora/blob/racon_denovo/src/denovo_discovery/racon.cpp#L38-L49 to how we call `racon`). I think one of main performance improvements we can do is to use [memfd](https://man7.org/linux/man-pages/man2/memfd_create.2.html) (i.e. create files in RAM) instead of actual files in the filesystem for `minimap2` and `racon`, as writing several files per locus to the disk is the bottleneck of discovery. This is noted as a TODO [here](https://github.com/rmcolq/pandora/blob/95fc951aa54d703d8f9c2cb53e228bdf82ae8f5b/src/denovo_discovery/racon.cpp#L141-L145). I have much of the code for `memfd` ready, `minimap2` works with it, `racon` does not though. It always error out [here](https://github.com/lbcb-sci/racon/blob/1a12aff289f9bc7523b968b41d0dcd24861acaf3/src/polisher.cpp#L85-L101) because it needs suffixes in its filepaths to recognise what type of file it is, but `memfd` files always look like `/proc/self/fd/63` or sth like that, e.g. no suffixes. I just need to change it to read `fasta` as default and I think will work... but got my time on `pandora` cut short...

The above would be a speed up on running `minimap2+racon` on a locus. We might also need a filtering on whether we even should run `minimap2+racon` on a locus, as you said. One way would be to check for drops of coverage, as the DBG denovo does, however this would not detect insertions on the ML sequence. If there are no new variants for a locus, we at least run `racon` just once and figure out it did not improve the sequence, and we finish the process... but we still run it...

iqbal-lab:
I suggest we move performance ideas to a new issue. Looks like we're ready to merge and add unit tests, or the other way round

leoisl:
Hey @mbhall88 , sorry, I just corrected two big bugs in the two most recent commits.

Bug 1: In [this commit](https://github.com/rmcolq/pandora/pull/299/commits/bb7b19208ab56f1f8e9dacda0f891c1b7e067fba) we started outputting SAM lines with flag 16, where the read maps as reverse complemented. However, when running discovery, we would just pull out reads where the flag was 0, i.e. I introduced a bug in this commit and `de novo` would just consider reads that mapped in the forward strand, effectively losing around half of the reads. This was fixed in this commit in [this commit](https://github.com/rmcolq/pandora/pull/299/commits/740f2a67c7d0ae1fad8a7259a761371df07f3165). This probably affects your results.

Bug 2: In [this commit](https://github.com/rmcolq/pandora/pull/299/commits/1b0afc0819a2a03f13685810c007aae06f231693), I changed the order of them SAM extra fields so that is more pleasant to look at, so that very long extra fields (i.e. the flanks and the PRG path) are the last fields. Unfortunately, `de novo racon` code would find the left and the right flank fields by using predefined indexes, and not looking by the `LF:Z` and `RF:Z` tag. This was just fixed in [this commit](https://github.com/rmcolq/pandora/pull/299/commits/9bb67c62c0a03549601c5d9737d776315ffaebdb). This severely impacts your results, as you basically lost the left and right flank, but this commit was on 27th October, maybe you did not pulled these changes when you ran `pandora`...

leoisl:
Finished memfd implementation (see https://github.com/rmcolq/pandora/issues/302) and improved the performance on some critical regions, which can yield improved runtime if `pandora discover` is ran with several threads. Just realised I have to do at least a few tests with different number of threads to see if the current code works as expected with several threads and how it scales with more threads

mbhall88:
[pandora-e164973.gz](https://github.com/rmcolq/pandora/files/9906156/pandora-e164973.gz)

mbhall88:
Ooops I assume the tip of this PR is not functional yet? With that binary I got the following error

```
[racon::createPolisher] error: file /proc/874910/fd/4 has unsupported format extension (valid extensions: .fasta, .fasta.gz, .fna, .fna.gz, .fa, .fa.gz, .fastq, .fastq.gz, .fq, .fq.gz)!
```

Will retry with the commit you suggested in #303 https://github.com/rmcolq/pandora/pull/299/commits/9bb67c62c0a03549601c5d9737d776315ffaebdb

mbhall88:
[pandora-9bb67c6.gz](https://github.com/rmcolq/pandora/files/9906278/pandora-9bb67c6.gz)

mbhall88:
> Ooops I assume the tip of this PR is not functional yet? With that binary I got the following error
> 
> ```
> [racon::createPolisher] error: file /proc/874910/fd/4 has unsupported format extension (valid extensions: .fasta, .fasta.gz, .fna, .fna.gz, .fa, .fa.gz, .fastq, .fastq.gz, .fq, .fq.gz)!
> ```
> 
> Will retry with the commit you suggested in #303 [9bb67c6](https://github.com/rmcolq/pandora/commit/9bb67c62c0a03549601c5d9737d776315ffaebdb)

Actually I get the same error with https://github.com/rmcolq/pandora/commit/9bb67c62c0a03549601c5d9737d776315ffaebdb as well

leoisl:
Sorry, forgot to say, I updated the racon fork to not error out on this,
please run git submodule update and it should work

leoisl:
Hey @mbhall88 , I am also starting to run this code on the paper data for benchmarking purposes. You can find a precompiled binary for the [latest commit](https://github.com/rmcolq/pandora/pull/299/commits/02f9ec954473d65bee16ca255dfeed72a94e71d6) attached.
[pandora-linux-precompiled-glibc.gz](https://github.com/rmcolq/pandora/files/9908820/pandora-linux-precompiled-glibc.gz)

leoisl:
Small update: started adding unit tests

mbhall88:
Are you happy with the testing I've done on this now @leoisl?

Anything else that needs to be done before merging?

leoisl:
Sorry for the delay, but I think you were on vacations this week anyway, right?

> Are you happy with the testing I've done on this now @leoisl?

Happy with your tests! Not happy with my lack of unit tests!



> Anything else that needs to be done before merging?

Well, there are some misc stuff like fixing the example script, changelog, etc... Should we do for this alpha release?

Issue #308, opened 2022-11-10
Remove `merge_index` command


leoisl:
`pandora` is able to merge different indexes into a single one: https://github.com/rmcolq/pandora/blob/a496870887a8c5d24c33caa051be0123d170a8d9/src/merge_index_main.cpp#L8-L21 . I never used this command since index building was optimised. I am now doing a bit of refactoring on pandora indexes (https://github.com/rmcolq/pandora/issues/306) and could I remove this command or do you think is still useful?

iqbal-lab:
not sure we need it? is this for merging two prgs of the same locus? or combining two different locus-prgs into one prg?

leoisl:
combining two different locus-prgs into one prg

iqbal-lab:
if no need given current index building, then fine i suppose. What if you wanted to index 1 million PRGs, might you parallelise and then merge results?

leoisl:
> What if you wanted to index 1 million PRGs, might you parallelise and then merge results?

That is an option, I wouldn't do that in roundhound though. Using 16 threads, we index 188k PRGs in 32 mins. Linear extrapolation would make us index 1M in ~3h, which is totally ok IMO. Indexing 10M PRGs might still be fine, might take ~30h with 16 threads, but there are still optimisations we could do and also increase nb of threads. But if you think we could potentially get into the order of 100M PRGs to index, then I think we should keep this command

iqbal-lab:
OK Fine I agree, let's get rid of it

Issue #307, opened 2022-11-10
Upgrading pandora required c++ version from c++11 to c++14


leoisl:
[Zipios](https://github.com/Zipios/Zipios) requires C++14 minimum, so I am upgrading pandora required C++ version from c++11 to c++14. Hope there is no issue with you, C++14 is already 8 years old.

iqbal-lab:
basically, can we compile on the cluster? i assume yes?

mbhall88:
Do you reckon we could get away with leapfrogging to c++17? From memory there's a lot of nice new features in that version?

leoisl:
Closed via https://github.com/rmcolq/pandora/pull/318

Issue #306, opened 2022-11-08
Refactor pandora index into a single, self-contained file


leoisl:
The `pandora` index is a set of files that need to live together: we need the `.prg.fa` file (the PRGs themselves), the `.prg.fa.kxx.wyy.idx` (the index itself), the `kmer_prgs/**/*.kxx.wyy.gfa` files (the kmer graphs, one file/PRG). We will probably add another file with the `PRG.SP_length` for each PRG (see https://github.com/rmcolq/pandora/issues/305). This structure is not ideal for very large PRGs, e.g. for our main use case in https://github.com/rmcolq/pandora/issues/304/, indexing will create ~1M `kmer_prgs/**/*.kxx.wyy.gfa` files, which makes it hard to distribute and move. Also when implementing https://github.com/rmcolq/pandora/issues/305, we will need to subsample the `.prg.fa` file to only the PRGs that remain in the index, as well as re-structure the the `kmer_prgs/**/*.kxx.wyy.gfa` files. This can quickly get messy.

The `kmer_prgs/**/*.kxx.wyy.gfa` could well be transformed into a single multi gfa file, as we don't randomly access these gfas, only sequential access. Then we have in the end a handful of files, which we can combine all into a single zip file, so that we can easily access each file fast, and also for compression.

`pandora index` shall create these `.zip` files and these will be consumed by the other `pandora` commands.  `pandora subindex` will read one `pandora` index (`.zip` file) and will create a subsampled `pandora` index (another `.zip` file). These files are self contained, so `pandora subindex` will just need to modify internal files inside this `zip` and every data the other `pandora` commands need to map reads to the index will also be in this zip file.

mbhall88:
I am all for the index being merged into a single file. This is also somewhat related to #276

leoisl:
After searching through the various C/C++ libraries that can work with `zip` files, I narrowed down to 3:
1. [Zipios](https://github.com/Zipios/Zipios): the easiest one to use by far. It abstracts lots of low-level details of handling archives, and gives us simple ways to read data from zip archives as a `C++` input stream, and simple ways to write data. It will be very simple to read and write to zip archives with this. My only hesitation is that is a new repo and not widely used;
2. [libarchive](https://github.com/libarchive/libarchive): heavily used and tested, very stable. However, it is a `C` library, so the interface is very `C`-ish. It mostly involves using low-level functions to read and write data, so it will be slower to implement what we need with this library. But I think is a nice backup plan if option 1 does not work;
[...]

PR #318, opened 2023-01-10
Changing pandora index format

leoisl:
This PR changes the pandora index from a set of files in a directory structure to a single, compressible and indexable `zip`
file (`pandora` indexes now have the suffix `.panidx.zip`). This is now the single file that is produced by the
`pandora index` command and is required as argument to all the other `pandora` commands. This index is self contained in 
the sense that it encodes all the information and metadata about it (e.g. which PRGs were used to create it, window and 
kmer size, etc). This new index provide the infrastructure for the next features and simplifies working with large 
reference pangenome collections, with a few million PRGs. These changes will be released as `pandora v0.11.0`.

Closes https://github.com/rmcolq/pandora/issues/308 https://github.com/rmcolq/pandora/issues/307 https://github.com/rmcolq/pandora/issues/306

## Breakdown of main changes
Sorry that this is another big PR, but half of the changes can be ignored as they are just updating the example data. Here is a breakdown of the main changes:
  * All the 29 files in the `example/` dir can be ignored, I just updated the example files to the latest version;
  * C++ version upgraded from 11 to 14 (`CMakeLists.txt`);
  * All `*_main(.h/.cpp)` files were changed to receive a pandora index (`*.panidx.zip`) instead of a fasta file containing PRGs;
  * `compare_main(.h/.cpp)`, `map_main(.h/.cpp)`, `discover_main(.h/.cpp)`, `seq2path_main(.h/.cpp)` all changed in a similar way to remove CLI parameters `-w` and `-k` (we now get this metadata from the index);
  * We have several refactorings to the `pandora` index implementation (`index(.h/.cpp)`);
  * We added a memory-efficient way to load PRGs when indexing, where we don't need to load all PRGs at once to index them, but just load on demand (`localPRG_reader(.h/.cpp)`);
  * Removed `merge_index` subcommand (`merge_index_main(.h/.cpp)`);
  * Added classes to deal with Zip files (`zip_file(.h/.cpp)`);

### Changed

- The `pandora` index changed from a set of files in a directory structure to a single, compressible and indexable `zip`
file (`pandora` indexes now have the suffix `.panidx.zip`). This is now the single file that is produced by the
`pandora index` command and is required as argument to all the other `pandora` commands. This index is self contained in 
the sense that it encodes all the information and metadata about it (e.g. which PRGs were used to create it, window and 
kmer size, etc). This new index provide the infrastructure for the next features and simplifies working with large 
reference pangenome collections, with a few million PRGs. This new index breaks backwards compatibility with previous 
`pandora` versions. The structure of this zip archive is as follows:
  * `_prgs`: The PRGs themselves used as input to create this index;
  * `_prg_names`: The names of the PRGs;
  * `_prg_min_path_lengths`: the length of the shortest path through each PRG;
  * `_minhash`: the minimizer hash data structure;
  * `_metadata`: metadata about the index (first line is window size, second is kmer size);
  * `*.gfa`: the several GFA files describing the minimizing kmer graph for each PRG;
- Minimum C++ standard upgraded from `C++11` to `C++14`;

### Removed
- Removed CLI parameters `-w` and `-k` from the following `pandora` subcommands: `compare`, `discover`, `map`,
`seq2path`;   
- Removed `merge_index` subcommand;

### Fixed
- Several refactoring to the `pandora` index implementation;

### Added
- A memory-efficient way to load PRGs when indexing, where we don't need to load all PRGs at once to index them, but 
just load on demand;


include/compare_main.h
```diff
@@ -29,11 +29,9 @@ using std::set;
 using std::vector;
 
 struct CompareOptions {
-    fs::path prgfile;
+    fs::path pandora_index_file;
```
mbhall88, line 32: Maybe just simpler as `index_file`? It's kind of obvious it's a pandora index no? Anyway, that's a very minor thing. Don't worry about changing it if you can't be bothered

CMakeLists.txt
```diff
@@ -12,20 +12,17 @@ HunterGate(
 
 # project configuration
 set(PROJECT_NAME_STR pandora)
-project(${PROJECT_NAME_STR} VERSION "0.10.0" LANGUAGES C CXX)
+project(${PROJECT_NAME_STR} VERSION "0.11.0" LANGUAGES C CXX)
```
mbhall88, line 15: We're okay to stay on 0.10.0 as we are still in alpha release, which basically says we are in the process of adding features for 0.10.0. Best not to skip 0.10.0 as we never did a stable release yet.

mbhall88 (changes requested):
Can't say I checked every bit of code, but love the changes. Especially getting rid of the need for k and w in map and compare and discover!

The one change I would recommend is relating to the version. We should stay on 0.10.0 but bump to 0.10.0-alpha.1 (see my comment in CMakeList.txt for more details)

CMakeLists.txt
leoisl: Done in https://github.com/rmcolq/pandora/pull/318/commits/a6f7db0ef50a3926d3c3e8ae971cb10941b103c7

include/compare_main.h
leoisl: Done in https://github.com/rmcolq/pandora/pull/318/commits/2631de2697d65def8622931117686db46f6b213b

leoisl:
Thanks for the comments, both were addressed! However, I am hesitant to pre-release this, as I've not tested on real data, and we're delaying writing unit tests for the new code for later. The main contribution of this PR is to provide infrastructure to the lazy loading of PRGs feature, which is essential to roundhound. This feature should be finished somewhere next week, and then I will test it with roundhound dataset, and will be more confident on pre-releasing it.

CMakeLists.txt
```diff
@@ -12,7 +12,7 @@ HunterGate(
 
 # project configuration
 set(PROJECT_NAME_STR pandora)
-project(${PROJECT_NAME_STR} VERSION "0.11.0" LANGUAGES C CXX)
+project(${PROJECT_NAME_STR} VERSION "0.10.0" LANGUAGES C CXX)
```
mbhall88, line 15: Just wondering, is it possible to make the version 0.10.0-alpha.1 here? I tried looking at the CMake docs and it wasn't super clear...

mbhall88 (approved):

mbhall88:
> However, I am hesitant to pre-release this, as I've not tested on real data, and we're delaying writing unit tests for the new code for later.

Sure. Do a pre-release whenever you think it makes sense (i.e. when there is a complete set/functionaly for the new features)

CMakeLists.txt
leoisl: I've tried this before and was unable to do it, but will retry just in case

leoisl: I didn't manage to do it, there is a 5-year open issue in CMake git repo asking for this feature, but hasn't been finished yet: https://gitlab.kitware.com/cmake/cmake/-/issues/16716 . So I am keeping the version as is I guess...

leoisl: Oh, we can actually use an extra number for the tweak version, so won't be able to set `0.10.0-alpha.1` but can set `0.10.0.1`.

leoisl: Just did this in https://github.com/rmcolq/pandora/pull/318/commits/524c7ae755e1a711c717ec3a9bf2aa3c5292399f , hope is ok. I am then merging this PR. If there is a problem, please tell me in slack or raise an issue.

PR #321, opened 2023-01-18
allow confidence to equal threshold

### LINKED ISSUE #320: Slight issue with checking minimum genotype confidence threshold
https://github.com/rmcolq/pandora/blob/565c1572f2c535db20c45062d359055d04a7659a/src/sampleinfo.cpp#L262-L263

the wording for this option on the command line is 

```
-G,--gt-conf INT            Minimum genotype confidence (GT_CONF) required to make a call [default: 1]
```

but we only check if the variant's gt conf is **greater** than the threshold. Given we are saying it is the minimum confidence required to make a call, it should be `>=`

The result of this is if I want to allow calls with gt conf 0 (equal coverage on ref and alt), I can't as the variable is a `uint16_t`, so setting to -1 will break the universe.

The other thing that may slightly complicate this is that `confidence` is a `double`... So checking for equality will require some care. We'll probably have to cast the thresold to a double and then do some kind of float equality comparison that allows for some difference (i.e. 0.0001)

mbhall88:
### Changed

- We now test whether the genotype confidence of a variant is greater than or equal to the threshold provided by `--gt-conf`. Previously we only tested if it was greater than. [closes #320]


src/sampleinfo.cpp
```diff
@@ -259,8 +259,10 @@ SampleInfo::get_genotype_from_coverage() const
         std::tie(index_of_max_likelihood, confidence, max_likelihood)
             = *index_and_confidence_and_max_likelihood_optional;
 
-        const bool satisfy_confidence_threshold
-            = confidence > genotyping_options->get_confidence_threshold();
+        const double threshold = genotyping_options->get_confidence_threshold();
+        const auto conf_equals_threshold = boost::math::relative_difference(confidence, threshold) < 1E-5;
```
leoisl, line 263: Just realised we already have implemented another way to check if doubles/floats are almost equal: https://github.com/mbhall88/pandora/blob/76c4cafef92015b204e5e5d350de3fc3c8f4ac51/include/Maths.h#L110-L118 . It uses what GoogleTest uses to check if doubles/floats are almost equal. I think we should discuss both methods and choose one. Source for the GoogleTest check can be found [here](https://github.com/google/googletest/blob/main/googletest/include/gtest/internal/gtest-internal.h#L334-L347). Boost float comparison docs [here](https://www.boost.org/doc/libs/1_81_0/libs/math/doc/html/math_toolkit/float_comparison.html) . If both methods are equivalent, I'd be slightly leaning towards the GoogleTest one, since it seems they carefully account for some edge cases (e.g. `nan`s), and how many units in the last place to tolerate when comparing doubles (tweaked to ordinary use on Intel CPUs, see [kMaxUlps](https://github.com/google/googletest/blob/ec25eea8f8237cf86c30703f59747e42f34b6f75/googletest/include/gtest/internal/gtest-internal.h#L274-L286)). However, I am going to help on predocs training now, and won't have time to continue reading or check this afterwards... I can discuss further tomorrow

leoisl (changes requested):

src/sampleinfo.cpp
mbhall88: Oh whoops, didn't realise we had a function already. I'll just use that then.  
Yeah, boost also has lots of edge case handling (see code [here](https://www.boost.org/doc/libs/1_62_0/boost/math/special_functions/relative_difference.hpp)). I think they're both effectively influenced by Knuth and [this blog post](https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/), which I would caution against. It sucked me in and I lost a whole morning and am now very scared of float comparison 😆

leoisl (approved):

Issue #330, opened 2023-06-12
Logging: new optimisations and 4-way results


leoisl:
# Description

This is just a logging for the new upcoming PR that will have 2 major changes in pandora:
1. Lazy loading of PRGs (improves RAM significantly in the plasmid/roundhound use case);
2. Do not keep read data (e.g. all minimiser hits each read has, all PRGs it maps to, etc...), but process this data as early as possible and release the memory. Improves RAM significantly in all cases;

And also minor changes:
1. Code cleanup, removing unused gene DBG and noise filtering modules;
2. If the best mapping of a read is to several graphs, we choose one at random (before was deterministic, so a single graph would get all the mappings)

# Results

The major changes should not impact results as they are just RAM improvements. The multimapping improvement should change the results slightly, but hopefully for better. To check if any breaking bug was added, we ran [this version](https://github.com/rmcolq/pandora/commits/b19d2683fb7a57eaefaacca64a0a436350f35c1f) against the [most updated prerelease](https://github.com/rmcolq/pandora/releases/tag/0.10.0-alpha.0) on the 4way pipeline. In general, the new version is slightly better precision-wise without denovo, and the old version is slightly better precision-wise with denovo. The differences are however small. RAM improvements are massive and will be detailed in a later post. This will enable pandora to be run with far less computational resources, and it will also speed up the next feature, which is running it on the cluster on hundreds of samples, so I think it is worth to merge these improvements.

# Details

Detailed 4-way results follow

Illumina filtered:
* This is the most important illumina results, with filters being cov: 5, strand bias 0.05, gaps 0.8 (the filters we used in the paper);
* Without denovo, the new version is slightly better precision-wise;
* With denovo, both versions are equivalent;

![image](https://github.com/rmcolq/pandora/assets/8397967/acc6208d-8051-4dfb-9847-b19089ac0204)

Illumina unfiltered:
* Shown here just for completeness. We should always apply lenient filters to pandora output...
* Without denovo, the new version is better for higher GT conf threshold. When GT conf is lowered, both versions converge to the same results;
* With denovo, v0.10.0-alpha.0 is slightly better precision-wise;

![image](https://github.com/rmcolq/pandora/assets/8397967/f23cc72e-ba16-4ec2-b07f-d78fb5a025ea)

Nanopore filtered:
* Without denovo, the new version is slightly better precision-wise;
* With denovo, v0.10.0-alpha.0 is slightly better precision-wise;

[...]

PR #331, opened 2023-06-15
Lazy loading

leoisl:
This PR mainly adds the lazy loading feature to pandora, which just loads a graph from the index if reads map to it in map/discover/compare. This reduces the RAM usage considerably in very large PanRGs, like the plasmid ones, where we have ~1M graphs. The RAM/runtime improvement compared to the current release is, WRT to the ESBL sample `SRR16977031`:

`v0.10.0-alpha.0` (baseline, current release)
RAM usage: 178.1 GB
Runtime: 130 minutes

This PR:
RAM usage: 124.5 GB (30% less RAM than baseline)
Runtime: 31.8 minutes (4 times faster than baseline)

It also adds a secondary, minor feature, where if a read maps equally well to several graphs, we choose one at random. The current code would systematically choose a single graph instead, thus missing the other mappings.

Warning: unit tests are not compiling, so we removed them for now. This increases our tech debt, we will have to solve this soon. We tested that our changes did not introduce any serious bugs on real data, see https://github.com/rmcolq/pandora/issues/330 for details

### DIFF
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -403,6 +403,8 @@ MinimizerHitClusters filter_clusters(
 void filter_clusters2(MinimizerHitClusters& clusters_of_hits,
     const uint32_t& genome_size)
 {
+    // TODO: this method is all commented out, tagging it for removal
+    // Currently let's just error out if we ever call it
     fatal_error("Not implemented");
     /*
     // Sort clusters by size, and filter out those small clusters which are entirely


leoisl:
Pinging @iqbal-lab @Danderson123  - should we wait/keep reviewing or should I merge?

iqbal-lab:
Sorry for this, but I am super busy helping michael, adrian, brice with their papers, and finding it hard to get space to review. Also,  a review from me is not the same  as from someone completely on top of the codebase. i vote merge

PR #335, opened 2023-07-04
Adds parameters dealing with auto-updating error rate and kmer model, and for gene filtering based on coverage

leoisl:
This PR:
1. Adds a parameter to deal with auto-updating error rate and kmer model. Before this PR, after processing sample `n`, the error rate and the model for kmer coverages would be automatically updated, based on the mapping of the reads from sample `n` to the PanRG. These updated values would then be used to map sample `n+1`. If sample `n` had some issues that caused its error rate to be unusual or its mapping to be unusually low, this could impact negatively sample `n+1` mapping. In general, auto-updating the error rate and kmer model would likely produce better results if the input data is well cleaned and has no serious issue. Unfortunately, this might not always be the case, and most users don't check well the input data. This is now an optional parameter, which defaults to not do this auto update:

```
  --auto-update-params        Automatically update error rate and kmer coverage model parameters based on the
mapping of the previous sample. It could potentially generate more accurate results in some very specific cases, 
first run should always have this flag off
```

2. Adds 3 parameters to control when a gene should be filtered out due to too low or too high coverage. This removes hard-coded values that were previously in `pandora`. This is basically done by looking at the mean coverage of the gene and comparing with the values given for these 3 params:

```
  --min-abs-gene-coverage FLOAT
                              Minimum absolute mean gene coverage to keep a gene. Given the coverage on the kmers of the maximum
likelihood path of a gene, we compute the mean gene coverage and compare with the value in this parameter. If the mean is
lower than this parameter, the gene is filtered out, e.g. if this parameter value is 3, then all genes with mean <3 will be filtered
out.
  --min-rel-gene-coverage FLOAT
                              Minimum relative mean gene coverage to keep a gene. This is a proportion, between 0.0 and 1.0. Given the
coverage on the kmers of the maximum likelihood path of a gene, we compute the mean gene coverage and compare with the
value in this parameter and the global coverage. If the mean is lower than the computed value, the gene is filtered out, e.g. if
this parameter value is 0.05, then all genes with mean < 5% of the global coverage will be filtered out.
  --max-rel-gene-coverage FLOAT
                              Maximum relative mean gene coverage to keep a gene. Given the coverage on the kmers of the maximum
likelihood path of a gene, we compute the mean gene coverage and compare with the value in this parameter and the global
coverage. If the mean is higher than the computed value, the gene is filtered out, e.g. if this parameter value is 10, then all
genes with mean >10 times the global coverage will be filtered out.


```


iqbal-lab:
I think we probaby need to increase this default value

    float max_relative_gene_coverage { 10 };

to 100 or something, as they might be on plasmids.

mbhall88:
I see this flag has also been added to `map`. I don't understand how one would auto-update the error rate for sample $n+1$ for `map`? Sorry if I am missing something obvious.

Also, my two cents: I feel it might be better to make auto-update the default - i.e. keep the current behaviour - and allow a user to turn it off, thus acknowledging their data isn't clean.

leoisl:
> I think we probaby need to increase this default value
> 
> ```
> float max_relative_gene_coverage { 10 };
> ```
> 
> to 100 or something, as they might be on plasmids.

Done in https://github.com/rmcolq/pandora/pull/335/commits/08354f57ebb55fad34490d8905bb12ddd040d5b0

leoisl:
> I see this flag has also been added to `map`. I don't understand how one would auto-update the error rate for sample n+1 for `map`? Sorry if I am missing something obvious.

Indeed it is not useful to update the error rate in `map` for now. In the near future we are planning to make `map` be able to map several samples instead of one, so it could be useful in this case. However, the kmer coverage model could be updated and is later used when getting the ML path in `map`: https://github.com/leoisl/pandora/blob/08354f57ebb55fad34490d8905bb12ddd040d5b0/src/map_main.cpp#L365-L368

Actually I just realised in `discover` we receive several samples, but we don't update the error model nor the kmer coverage model at all (i.e. we don't have a call to `estimate_parameters()`). Do you think this is some sort of bug? Should we add this to `discover`?

mbhall88:
> Actually I just realised in discover we receive several samples, but we don't update the error model nor the kmer coverage model at all (i.e. we don't have a call to estimate_parameters()). Do you think this is some sort of bug? Should we add this to discover?

Hmmm I don't know. `discover` has changed a lot since I was last working on it. I think last I was working on it (before the racon change) it wasn't able to take more than one sample so this wouldn't have been a problem no? But if we're going to allow multiple samples, probably better to mirror what `map` and `compare` do. Though, we run racon on every PRG anyway right, so not sure it would make a difference? Would only make a difference if we were relying on kmer coverage in order to trigger running racon (which might be wise to do at some stage)

leoisl:
> I think last I was working on it (before the racon change) it wasn't able to take more than one sample so this wouldn't have been a problem no?

No, it wouldn't.



> But if we're going to allow multiple samples, probably better to mirror what `map` and `compare` do. Though, we run racon on every PRG anyway right, so not sure it would make a difference? Would only make a difference if we were relying on kmer coverage in order to trigger running racon (which might be wise to do at some stage)

Updating the kmer coverage model right now makes no difference on discovery, but updating the error rate does make, as `discovery` will now not use a single fixed error rate, but an error rate that is auto updated after processing a sample. I chose to do this to indeed keep consistency with the other 2 commands, implemented in https://github.com/rmcolq/pandora/pull/335/commits/e827462bd2d887a06d29483036ac9db1f0f77d36 . The option `--dont-auto-update-params` was also added to discovery, in case you want to fallback to the previous behaviour.

leoisl:
Hey @iqbal-lab , @mbhall88 , @Danderson123 I think this is ready to be merged, please take a look and state your corrections or further reviews, if any.

include/compare_main.h
```diff
@@ -41,7 +41,11 @@ struct CompareOptions {
     bool output_vcf { false };
     bool illumina { false };
     bool clean { false };
+    float min_absolute_gene_coverage { 3.0 };
```
iqbal-lab, line 44: what is absolute gene coverage? number of reads on the gene, or the mean depth?

include/denovo_discovery/discover_main.h
```diff
@@ -29,7 +29,11 @@ struct DiscoverOptions {
     bool illumina { false };
     bool clean { false };
     bool binomial { false };
+    bool do_not_auto_update_params { false };
     uint32_t max_covg { 600 };
+    float min_absolute_gene_coverage { 3.0 };
+    float min_relative_gene_coverage { 0.05 };
+    float max_relative_gene_coverage { 10 };
```
iqbal-lab, line 36: we define the same 3 vars to the same defaults in 2 files. Feels to me like we're going to accidentally modify a default value in one place and not the other, and end up confused. i think these consts should be the same for discover and compare?

include/map_main.h
```diff
@@ -44,7 +44,11 @@ struct MapOptions {
     bool illumina { false };
     bool clean { false };
     bool binomial { false };
+    bool do_not_auto_update_params { false };
     uint32_t max_covg { 300 };
+    float min_absolute_gene_coverage { 3.0 };
+    float min_relative_gene_coverage { 0.05 };
+    float max_relative_gene_coverage { 100 };
```
iqbal-lab, line 51: oh look we define them here too

src/compare_main.cpp
```diff
@@ -89,12 +89,62 @@ void setup_compare_subcommand(CLI::App& app)
...
+    compare_subcmd
+        ->add_option(
+            "--min-rel-gene-coverage", opt->min_relative_gene_coverage,
+            "Minimum relative mean gene coverage to keep a gene. This is a proportion, between 0.0 and 1.0. "
+            "Given the coverage on the kmers of the maximum likelihood path of a gene, we compute "
+            "the mean gene coverage and compare with the value in this "
+            "parameter and the global coverage. If the mean is lower"
+            " than the computed value, the gene is filtered out, e.g. if this parameter value is "
+            "0.05, then all genes with mean < 5% of the global coverage will be "
+            "filtered out.")
```
iqbal-lab, line 129: ah i see, the internal vars directly reflect user specified vals. Which si fine, i just wonder about the defaults. it's a small thing

src/denovo_discovery/discover_main.cpp
```diff
@@ -88,12 +88,62 @@ void setup_discover_subcommand(CLI::App& app)
...
 
+    discover_subcmd
+        ->add_option(
+            "--min-abs-gene-coverage", opt->min_absolute_gene_coverage,
+            "Minimum absolute mean gene coverage to keep a gene. Given the "
+            "coverage on the kmers of the maximum likelihood path of a gene, we compute "
+            "the mean gene coverage and compare with the value in this "
+            "parameter. If the mean is lower than this parameter, "
+            "the gene is filtered out, e.g. if this parameter value is "
+            "3, then all genes with mean <3 will be filtered out.")
```
iqbal-lab, line 114: if this value is here; my guess is that many reads get annotated with a gene in the SAM file during mapping, but then pandora looks across all the information, and then decides whether a gene is truly present. If so, and pandora has decided the gene is not there, @Danderson123 might need to remove those gene annotations from his reads.

src/estimate_parameters.cpp
```diff
@@ -354,5 +357,12 @@ uint32_t estimate_parameters(std::shared_ptr<pangenome::Graph> pangraph,
     for (auto& node : pangraph->nodes) {
```
iqbal-lab, line 357: could we have a brief chat about what this is fdoing?

iqbal-lab:
Happy to merge after the above is discussed.

include/compare_main.h
leoisl: gene coverage is always mean depth. Then we have minimum absolute gene coverage, a flat value that the mean gene coverage needs to be kept, and the min relative gene coverage, a proportion WRT the global coverage

include/map_main.h
leoisl: Yeah, actually all CLI defaults are duplicated... will refactor all of them, not just these 3

src/denovo_discovery/discover_main.cpp
leoisl: Yeah, AFAIK @Danderson123 just processes the SAM file, and the SAM file is just read mapping, there is no FP filtration, FP gene filtration happens later

src/estimate_parameters.cpp
leoisl: Yep, what exactly? This `for` or the changes in this file?

iqbal-lab: basically i dont understand what the code is doing (before or after your change)

src/denovo_discovery/discover_main.cpp
iqbal-lab: ok so this is not relevant to this PR, but is a reminder to @Danderson123 that if, when he takes the pandora SAM, he also parses the gene presence file, he will get the clean filtered list of genes present, and can make a dictionary of them , and when parsing the SAM, can ignore genes on reads which are not in the dictionary

iqbal-lab:
ok to merge

